PR #38, opened 2015-07-14
Working Travis Configuration

lumip:
I've finally been able to come up with a configuration for Travis that works.
I've created a preliminary setup.py, which is used by Travis to run tests via the "setup.py test" command.

To ensure that imports in test modules work in any case (i.e. Travis and manual execution in eclipse) I had to change the way imports had been: Every test module now adds the src-folder to sys.path (instead of the concrete subfolder, e.g., src/pulses). Accordingly, imports in test modules must be made relative to that, e.g. import pulses.Parameter. This is the only solution I could come up with such that tests run in all required cases.

I also had to (temporarily) add the typing module from my Python 3.5.0b2 directory since Travis currently only features Python 3.5.0a2 which seems to no include it.

Any remarks or proposals for changes? @ekammerloher @j340m3

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@
 *__pycache__*
 doc/*
 .*project
+.eggs*
+*.egg-info*
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,11 @@ language: python
 python:
   - "nightly"
 # command to install dependencies
-#install: "pip install -r requirements.txt"
-# command to run tests
-script: sh -c 'cd tests && python3 run_tests.py'
+install:
+    - "pip install pyflakes"
+    - "pip install coverage"
+    - "pip install coveralls"
+script:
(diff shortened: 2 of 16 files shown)


ekammerloher:
I think it is a good design change to use a setup.py, since it conforms to usual practice. It will be easier to include dependencies and make our toolkit pip-installable.

Tools like Travis and Coveralls might not add too much value in the short term, but will help advertise our project as a serious development effort to the community in the long term and are good to keep track of development over time.

In this sense I also enabled coveralls.io and modified the README.md

lumip:
Before we merge, I want to know if @j340m3 is okay with the changes to the imports since I've basically reverted some of his recent work with that.

j340m3:
If everything works, this looks fine for me. The changes to the import (i.e. try and except around it) where hotfixes from my side and not intended to be long term code.

Issue #53, opened 2015-08-07
Change docstring style for Sphinx documentation


pbethke:
We decided to follow the [Google Python Style](https://google-styleguide.googlecode.com/svn/trunk/pyguide.html#Comments) for this. Another great example covering an exhaustive range of cases can be found [here](http://sphinxcontrib-napoleon.readthedocs.org/en/latest/example_google.html).

I have already adapted a large part of the comments and the two things that remain to be done are:
- Someone look over the changes
- Read the style guidelines for future docstrings

This change mostly concerns @lumip and @j340m3, but @qutech/bluhm-group should be aware of this.

PR #54, opened 2015-08-07
Make the trivial changes to Google style docstrings

pbethke:
Relates to #53.

### DIFF
--- a/src/pulses/BranchPulseTemplate.py
+++ b/src/pulses/BranchPulseTemplate.py
@@ -11,7 +11,7 @@ from .PulseTemplate import PulseTemplate, MeasurementWindow
 logger = logging.getLogger(__name__)
 
 class BranchPulseTemplate(PulseTemplate):
-    """!@brief Conditional branching in a pulse.
+    """Conditional branching in a pulse.
     
     A BranchPulseTemplate is a PulseTemplate
     with different structures depending on a certain condition.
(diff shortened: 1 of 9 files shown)


lumip:
This seems fine. However, in the docstring of the ParameterDeclaration constructor, the information that the arguments are keyword arguments are lost. The example referenced in #53 states that
"If \*args or \*\*kwargs are accepted,
    they should be listed as ``*args`` and ``**kwargs`` [in the 'Args:' section]."

pbethke:
Why do you use **kwargs anyway instead of naming them?

This whole block is a bit weird:
```python
def __init__(self, **kwargs) -> None:
    # Docstring
    super().__init__()
    self._min_value = None
    self._max_value = None
    self._default_value = None
    for key in kwargs:
        value = kwargs[key]
        if isinstance(value, numbers.Real):
            if key == "min":
                self._min_value = value
            elif key == "max":
                self._max_value = value
            elif key == "default":
                self._default_value = value
        else:
            raise ValueError("{0} is not a valid argument.".format(key))
        else:
            ... # Assertions
```

the only values in **kwargs that are ever going to matter are ```min, max, default```. You check if they are present and replace them by their default value otherwise. Why not just use

```python
def __init__(self, min=None, max=None, default=None) -> None:
    # Assertions, make sure it's a real number, etc.
```
instead?

lumip:
The reasoning behind this was that the parameters are all optional. Maybe a user only wants to specify a default value but not min and max. It is then more convenient and descriptive (in my opinion) to just type `ParameterDeclaration(default=0.4)` instead of `ParameterDeclaration(None, None, 0.4)`. 
After all, such a simplification for the user is the essential benefit I see in keyword arguments.

pbethke:
keyword parameters in Python are always optional and unordered, so ```ParameterDeclaration(default=4.0)``` is perfectly fine.

lumip:
Ah, I see. I'll adapt the parameter list to follow your suggestion.

PR #61, opened 2015-08-21
Dev patrick

pbethke:
A lot of bugfixes for tablepulsetemplate, including a rewrite of the interpolation stuff. I don't want it lying around until merging becomes difficult


lumip:
- Before merging, we should resolve the failing tests. This essentially means including numpy in the setup.py dependencies. I'm currently working on this (and installing numpy on my machine).
- We should decide whether we want to use numpy arrays or regular arrays/lists and do so consistently.
- I don't entirely agree with the serialization. The current implementation restricts us to JSON-only and I might not hurt to construct this to be more flexible, allowing to add further serialization targets without adding to_xyz() methods for each.

pbethke:
OK,
* I found [this gist](https://gist.github.com/dan-blanchard/7045057) outlining how to do numpy/scipy builds using miniconda. That won't help us with python 3.5, but enable us to test all python versions supported by anaconda.
* I agree that we should use consistent types. The numerical *rendered* pulse should probably be numpy arrays and everything template-related just built-in data types (see next point).
* Let's not merge the serialization things for now. I'll just branch it off. It was just a proof-of-concept after all, although i don't think other serialization layers work without specific handlers (except for the very limited python pickle and variants thereof). The ```.to_json``` just returns a representation of the class's data in terms of *relatively* language-agnostic primitive data types which might be useful for other serialization methods as well.
* I have a clever reason for this that only becomes important once we concatenate pulses: The total times of subtemplates might not necessarily agree exactly with the sample timing, so that there might be an offset of less that one clock-tick. The waveform will be more accurately represented by sampling at those offset clock times.
* We can definitely do that.

Issue #80, opened 2015-10-01
Handle qctoolkit as an package


terrorfisch:
Handle qc-toolkit as a package and 'pulses','utils' etc as subpackages.

terrorfisch:
While working on making the package install work and making the tests use the actually installed package and not the src dir, I ran into a few problems.

1. I found that all modules are spammed with names like 'Tuple', 'Parameter' and everything else that is imported. This is not desirable and makes using the auto-complete function a pain in the arse.
2. I do not think it is reasonable to put each PulseTemplate in a different module. But I guess we have to refactor the code organization anyway.

To solve these we need some conventions for imports. For the namespace cleanup I see two options: import with underscore or delete classes/modules at the end of the file. I am clearly in favor of the underscore variant:
'''
#like this
import typing as _type

#or like this
from abc import ABCMeta as _ABCMeta,

class foo(metaclass = _ABCMeta):

'''

terrorfisch:
While working on making the package install work and making the tests use the actually installed package and not the src dir, I ran into a few problems.

1. I found that all modules are spammed with names like 'Tuple', 'Parameter' and everything else that is imported. This is not desirable and makes using the auto-complete function a pain in the arse.
2. I do not think it is reasonable to put each PulseTemplate in a different module. But I guess we have to refactor the code organization anyway.

To solve these we need some conventions for imports. For the namespace cleanup I see two options: import with underscore or delete classes/modules at the end of the file.
```
#1: like this
import typing as _type

#2: like this
from abc import ABCMeta as _ABCMeta

#3: like this
import logging

logger = logging.getLogger(__name__)

class foo(metaclass = _ABCMeta):
    def __init__(self, bar: _type.Set[str])
        pass
del logging
```
I'm in favour of variant 1.

lumip:
1. I think I don't understand the problem here. If modules use Tuples and Parameters, it makes sense that these are defined therein after the import? 
Or is the problem that these symbols are exported by modules that import them? We could (and should in any case) eliminate that by restricting the exported symbols using __all__ in each module.

2. That bothers me, too. On the one hand, I don't want the module files to get insanely large and I'm always a fan of the one class one file paradigmn. On the other hand, this makes imports awkward in Python. Currently, I think both solutions are ugly in their own way, but I tend to favor grouping more similar classes in bigger files at the moment.

pbethke:
[...]

PR #81, opened 2015-10-01
Handle qctoolkit as a package itself.

terrorfisch:
Update of the setup script as written in #80

@lumip @pbethke please have a look if this makes sense.


lumip:
Seems fine but I would recommend waiting before merging until all points mentioned in #80 are resolved.

pbethke:
closed until the issues are resolved

PR #84, opened 2015-10-09
Matlab engine wrapper

terrorfisch:
I wrote some managing utilities which automatically connect to an existing matlab session/engine with the correct name if you import the subpackage.


terrorfisch:
Wrong target branch

Issue #132, opened 2016-10-29
Make measurement windows (re-)nameable


terrorfisch:
The user wants to distiguish and group the measurements into groups that i.e. share a common physical parameter. PulseTemplates should be able to rename the measurement windows of their children.

Example:
The user defines a parametrizerd SequencePulseTemplate P(x) which consists of the TablePulseTemplate 'T(x)' and a measurement pulse M: 'P(x) = [T(x) M]'
M contains a measurement window with the name 'mw_name'
The user wants to repeat P while ramping x from 1 to 100 and give each measurement_widow another name:

``` python
S = [ P(x_i,{'mw_name' : 'mw_{}'.format(i)}) for i,x_i in enumerate(x) ]
```

This pulse S is than repeated 300 times.

Measurement windows with the same name can than be given to the currrently used acquisition library as a mask.

terrorfisch:
Implemented with 230bc33490d11117a38c31999217f3d1983d5fb0

PR #133, opened 2016-11-22
Towards experiment

terrorfisch:
They now have an identifier that can be remapped for subtemplates
They end up in the EXEC instruction

Implements #132


terrorfisch:
This branch works now on a simple setup but has many unclean fixes as I hacked some things together for the backend which are not easyly testable because they are not granular and/or the module cannot be loaded without hardware. I have a new branch to improve test integration of as many features as possible without the hardware

PR #135, opened 2017-03-09
Make unit tests great again

terrorfisch:
Lets see how much work this is...


terrorfisch:
Next step: increase converage

Issue #145, opened 2017-10-19
ForLoopPT duration calculation is broken


terrorfisch:
The duration calculation of the ForLoopPulseTemplate is broken.

terrorfisch:
Fixed in #146

PR #146, opened 2017-10-19
Issues/145 foor loop duration

terrorfisch:
Fix bug #145 by using sympy.Sum for duration calculation.

sympy.Sum does not like the default lambdify/numpy evaluation of Expressions. I had to implement a flag to turn this of and a method to propagate that flag easily. This was done by adding basic math methods to the ```Expression``` class.


terrorfisch:
Old master state was broken. Will fix in new commit

Issue #167, opened 2018-03-06
Measurements to non atomic PT


qutech-lab:
It would be good, to allow measurements within other templates than just atomic pulse templates.
Especially RepetitionPT.
(Frederik Wangelik)

terrorfisch:
Currently measurements are bound to waveforms.

For now I will let the non-atomic pulse templates inject the measurements into the first waveform/EXEC instruction but there has to be a better way...

terrorfisch:
Does not work as the RepetitionPT currently has no way to access the instructions generated by the body.

terrorfisch:
Maybe a new MEAS instruction will seal the deal.

terrorfisch:
This would also be a good oportunity to prepare the comeback of the CHAN instruction and non-atomic multi channel pulse templates.

terrorfisch:
Implemented with merge #168 

I close this now. Please re-open if other PulseTempaltes are desired.

terrorfisch:
Need measurements in ForLoopPT

terrorfisch:
Implemented in #172

PR #168, opened 2018-03-06
Issues/167 non atomic pt mesurements

terrorfisch:
See issue #167


terrorfisch:
Allows declaring measurements for ```RepetitionPT``` and ```SequencePT``` and allows easier implementation for other non-atomic PulseTemplates. Moves measurements from ```Waveform``` to a seperate intruction ```MEASInstruction```

Issue #176, opened 2018-03-07
Inaccurate time calculations for long pulses


terrorfisch:
For very long pulses there might be numerical errors when calculating the pulse duration in samples.

For now I will use ```fractions.Fraction``` as a type for times. If we need the performance there is the possibility to switch to ```gmpy2.mpq```.

terrorfisch:
Should be solved now.

terrorfisch:
`gmpy2.mpq` gives a speedup of a few seconds for very long and complicated pulses.

PR #178, opened 2018-03-08
Issues/176 time type fraction

terrorfisch:
Solve #176 by using fractions.Fraction. Type can be changed easily as long as it implements ```__float__```


terrorfisch:
Additional tests that I will push later show that this probably won't break anything.

PR #190, opened 2018-03-15
add convenience function to convert expression to float

peendebak:
This allows one to write `float(pulse.calculate_duration())`

@terrorfisch @lucblom


terrorfisch:
I do not understand why this breaks the tests.

Anyways, I am currently rewriting parts of the expression stuff. For what do you need ```float(pulse.duration)```?

terrorfisch:
I found out why it breaks the test. Python converts to float instead of using __radd__.

PR #227, opened 2018-04-03
ParameterConstrainer.parameter_constraints now returns a set instead of an ordered list.

lumip:
Also: moved ParameterContrainerTests from measurement_tests.py to parameter_tests.py


tests/pulses/multi_channel_pulse_template_tests.py
```diff
@@ -319,8 +319,8 @@ def test_build_waveform_none(self):
         self.assertIsNone(wf)
 
     def test_deserialize(self):
-        sts = [DummyPulseTemplate(duration='t1', defined_channels={'A'}, parameter_names={'a', 'b'}),
-               DummyPulseTemplate(duration='t1', defined_channels={'B'}, parameter_names={'a', 'c'})]
+        sts = [DummyPulseTemplate(duration='t1', defined_channels={'A'}, parameter_names=['a', 'b']),
+               DummyPulseTemplate(duration='t1', defined_channels={'B'}, parameter_names=['a', 'c'])]
```
terrorfisch: Why this?

qctoolkit/pulses/function_pulse_template.py
```diff
@@ -126,7 +126,7 @@ def get_serialization_data(self, serializer: Serializer) -> Dict[str, Any]:
             expression=self.__expression,
             channel=self.__channel,
             measurement_declarations=self.measurement_declarations,
-            parameter_constraints=[str(c) for c in self.parameter_constraints]
+            parameter_constraints={str(c) for c in self.parameter_constraints}
```
terrorfisch: JSON does not support sets. This is converted into a list anyways.

terrorfisch:
I don't know if I a happy with this. Keeping the order might make sense from a UI perspective.

tests/pulses/multi_channel_pulse_template_tests.py
lumip: I actually can't tell. Reverted in most current commit.

lumip:
@terrorfisch I think because parameter constraints are a set conceptually, this should be reflected by the data structures holding them.

If I recall correctly, your argument was that users might prefer a certain order and thus it would make sense to return the constraints in the order they arrived. I would argue that I might not really see a point there.. if the parameter constraint set is small, order doesn't really matter for comparing. If it's large, one shouldn't manually compare but use set operations to check for existent/non-existent constraints (or subsets of them).

Also, different people might have different preferences as regarding to constraint order, making a pulse template "unportable" between persons (at least annoying one of them).

PR #239, opened 2018-04-16
Changed SequencePT.requires_stop to conform to old semantics

lumip:
The requires_stop method is used by the Sequencer and should only return False if the currently processed PT cannot be sequenced. For the SequencePT this is only the case when parameter constraints cannot be evaluated, and does not depend on subtemplates.


lumip:
No more conflicts, please merge @terrorfisch

lumip:
With the new sequencing procedures, this becomes obsolete (cf. #286 and related)

Issue #231, opened 2018-04-09
Remove external_parameters keyword argument from SequencePulseTemplate interface


terrorfisch:
`SequencePulseTemplate.__init__` currently has an optional keyword argument `èxternal_parameters`. If provided, the external parameters are checked to be equal to it.

This feature has seen no use in real world applications and is buggy. I suggest removing it.

lumip:
I have done so in the pull request above. We could also think about setting the default value for allow_partial_parameter_mapping in MappingPulseTemplate to True and thus further relax the strictness of mapping specifications.

PR #243, opened 2018-04-16
Deprecating external_parameters (SequencePT)

lumip:
As specified in #231 .


lumip:
@terrorfisch consider merging pls. I won't deal with coverage decrease that arises from removing tests for deprecated features here

Issue #222, opened 2018-03-27
Examples 9* out of date


thangleiter:
The first cell of the Detailed Sequencing Walkthrough notebook already uses some deprecated (?) features. This seems mostly centered around `LoopPulseTemplate`. Moreover, importing `SoftwareCondition` works only as 
```python
from qctoolkit.pulses.conditions import SoftwareCondition
```

thangleiter:
The same goes for Example 98 (also out of date in declaring measurements as far as I can see).

terrorfisch:
All 9* examples do not work or are outdated and are therefore not in the documentation(`doc\source\examples\examples.rst`). I kept them in the repository so one does not have to dig through old commits to find them.

It would be nice if you could make them work.

lumip:
Examples starting with 9* now either have a more up-to-date counterpart in the non-9* examples or are there to document outdated features. "Fixing" them or trying to "make them work" therefore does not really make sense. I propose keeping them for reference purposes but leave them unchanged.

PR #246, opened 2018-04-23
Examples update

thangleiter:
Updated examples 9* to comply with the current syntax except for 99 (don't understand the error) and the second example in 95 (uses `BranchPulseTemplate`; is there still a point to this one?) and added a notebook going into more detail with parameter mappings and expression vectors (94).


terrorfisch:
Partially tackles #222

PR #249, opened 2018-04-24
Ignore RepetitionPT body if repetition count <= 0

### LINKED ISSUE #248: RepetitionPulseTemplate sequences body if repetition count is null
The adding of the repetition instruction and sequencing of the body should only happen if the repetition count is larger null

terrorfisch:
Fixes #248


lumip:
Changes look good to me. I added tests and will merge.

PR #255, opened 2018-05-07
Fix bug in make_compatible when doing a complete unroll of a Loop

terrorfisch:


terrorfisch:
This or some earlier change broke the legacy compability test.

terrorfisch:
The tests rely on TaborSegment behaviour which changed

Issue #250, opened 2018-04-24
Serializer deserializes the same pulse twice


superpostition:
pulses.SequencePT() raises

Exception: Identifier '<...>' assigned twice.

if subplustemplates consist of same subpulstemplates.

terrorfisch:
The problem occurs because the serializer deserializes the same pulse/file twice. This means there are two distinct objects with the same identifier which leads to an error when one wants to save them again.

terrorfisch:
Changed priority as this is partially fixed now.

PR #259, opened 2018-05-16
Temporary fix for using the same pulse several times

terrorfisch:


terrorfisch:
Temporary fix for #250

PR #263, opened 2018-05-23
Complete serialization overhaul

terrorfisch:
The aim of this overhaul is to make Serialization more future proof and lay the groundwork for easier pulse management. Also solves #250 completely.

- `dict` interface where ever it makes sense (`StorageBackend`s i.e.)
- Serialization based on the extenability of `json.JSONEncoder` and `json.JSONDecoder`. This leads to much less code in each `PulseTemplate` but requries that references are marked explicitly.
- A PulseStorage class that also helps organizing pulses (draft)
- A central registration for all Serializables to avoid identifier duplication if not explicitly disabled

Some of these are backward incompatible and need to be discussed in detail.


terrorfisch:
Maybe introduce a new method `to_dict` for the new serialization interface and just dereprecate the old one.

lumip:
I also tried to think about a transition period where the old routines are still available but deprecated but I don't really want to rename the relevant methods in the new implementations (and can't in the old one) (to_dict is much less precise a name then get_serialization_data() )... it would be nice if Python would support method overloading but meh

lumip:
We should be about to wrap this up. Can merge into master after incorporating pull requests
#307 #306 #304  and thus closing issues #272, #294, #296, #298, #301
and possibly merging #300 to address #299 . 
The remaining issue #274 is not required for merge into master and can/will be addressed in the future.

terrorfisch:
I think we can merge now.

This will:
close #250 and 
close #263

lumip:
let's go

PR #266, opened 2018-06-07
backward-compatible implementations of get_serialization_data and deserialize

lumip:
All PulseTemplate classes now have get_serialization_data and deserialize methods that are conformant with the new serialization routines (from the base branch) but also maintain compatability to the old routines for a transition period. Tests cover both (old and new) ways.

First step to fully integrate the changes into master. 
Still coming up (in separate pull requests):
- Fixes to tests that currently still use the old serialization routines
- Script to automatically convert old serialized pulses into new format


lumip:
This is done and should be merged so that follow up changes can then also be integrated..

PR #288, opened 2018-07-05
Fix (global) pulse registry

terrorfisch:
- Rename pulse registration to registry
- Include in each pulse init method
- Trigger garbage collection on each pulse initialization.
- Update serialization tests to use dedicated registry dicts

### DIFF
--- a/qctoolkit/serialization.py
+++ b/qctoolkit/serialization.py
@@ -17,6 +17,7 @@ import tempfile
 import json
 import weakref
 import warnings
+import gc
 
 from qctoolkit.utils.types import DocStringABCMeta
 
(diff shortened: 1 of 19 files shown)


qctoolkit/serialization.py
```diff
@@ -351,18 +352,21 @@ def __init__(self, identifier: Optional[str]=None, registration: weakref.WeakVal
...
 
         if identifier == '':
             raise ValueError("Identifier must not be empty.")
         self.__identifier = identifier
 
-        if identifier and registration:
-            if identifier in registration:
+        if identifier and registry is not None:
+            # trigger garbage collection to "simulate" RAII
+            gc.collect(2)
```
lumip: This increases the time required to run all tests on my machine by a factor of ~3 (~20 to ~60 secs). Why exactly is this necessary?

terrorfisch: Because garbage collection is not deterministic and there might me a pulse with the same name that has no references left but isn't collected yet.

PR #304, opened 2018-07-12
Serialization conversion script

### LINKED ISSUE #296: Provide methods to easily convert old stored pulses
so that they can be loaded using the new serialization routines

lumip:
Methods convert_stored_pulse_in_storage() and convert_pulses_in_storage() to convert a specified or all pulses from a source storage where the are stored by the old serialization routines to a new destination backend using the new serialization routines.

Closes #296


qctoolkit/serialization.py
```diff
@@ -609,7 +645,7 @@ def deserialize(self, representation: Union[str, Dict[str, Any]]) -> Serializabl
         if isinstance(representation, str):
             if representation in self.__subpulses:
                 return self.__subpulses[representation].serializable
-        
+
```
terrorfisch: Please cleaun up to avoid merge conflicts

qctoolkit/serialization.py
```diff
@@ -141,6 +150,21 @@ def delete(self, identifier):
         except FileNotFoundError as fnf:
             raise KeyError(identifier) from fnf
 
+    def list_contents(self) -> Set[str]:
+        contents = set()
+        for dirpath, dirs, files in os.walk(self._root):
```
terrorfisch: `os.walk` goes through the complete tree. We only return top level jsons in `FilesystemBackend.get`

qctoolkit/serialization.py
```diff
@@ -91,6 +92,14 @@ def delete(self, identifier: str) -> None:
     def __delitem__(self, identifier: str) -> None:
         self.delete(identifier)
 
+    @abstractmethod
+    def list_contents(self) -> Set[str]:
```
terrorfisch: I suggest naming the method `keys`

qctoolkit/serialization.py
lumip: Which is why there's a line 159 which aborts the iteration after the first level. The reason to still use os.walk() was that it conveniently separates dirs and files and gives me the base dirpath as well in one single call.

qctoolkit/serialization.py
lumip: I think list_contents() is a more fitting name for what is going on here since the method returns a list of the (available) contents of the storage backend. keys() is a much more generic name and restricts to a plain dictionary "view" of the interface without regarding the context.. I guess it's more "pythonic" though.. ultimately, I don't really care

terrorfisch: I think sticking to `list_contents` is better as we are not returning a view. The return value of a keys method should update if the content of the mapping changes.

PR #305, opened 2018-07-12
Issues/301 pulse storage as default issue

lumip:


terrorfisch:
`JSONSerializableEncoder` and `JSONSerializableDecoder` can be used with any Mapping

Issue #302, opened 2018-07-12
Travis tests failing


lumip:
The following tests are suddenly failing on travis across all branches without an apparent change to the tested functionality in question. Problems with sympy dependency?

```
======================================================================
ERROR: test_indexing (tests.expression_tests.ExpressionScalarTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/qutech/qc-toolkit/qctoolkit/expressions.py", line 40, in _parse_evaluate_numeric_arguments
    return {v: eval_args[v] for v in self.variables}
  File "/home/travis/build/qutech/qc-toolkit/qctoolkit/expressions.py", line 40, in <dictcomp>
    return {v: eval_args[v] for v in self.variables}
KeyError: 'a[i]'
The above exception was the direct cause of the following exception:
Traceback (most recent call last):
  File "/home/travis/build/qutech/qc-toolkit/tests/expression_tests.py", line 137, in test_indexing
    self.assertEqual(e.evaluate_numeric(**params), 2 * 2)
  File "/home/travis/build/qutech/qc-toolkit/qctoolkit/expressions.py", line 62, in evaluate_numeric
    parsed_kwargs = self._parse_evaluate_numeric_arguments(kwargs)
  File "/home/travis/build/qutech/qc-toolkit/qctoolkit/expressions.py", line 42, in _parse_evaluate_numeric_arguments
    raise ExpressionVariableMissingException(key_error.args[0], self) from key_error
qctoolkit.expressions.ExpressionVariableMissingException: Could not evaluate <c*a[i]>: A value for variable <a[i]> is missing!
======================================================================
ERROR: test_eval_sum (tests.utils.sympy_tests.EvaluationTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/qutech/qc-toolkit/tests/utils/sympy_tests.py", line 251, in test_eval_sum
    result = self.evaluate(expr, parameters)
  File "/home/travis/build/qutech/qc-toolkit/tests/utils/sympy_tests.py", line 226, in evaluate
    return evaluate_lambdified(expression, variables=list(variables), parameters=parameters, lambdified=None)[0]
  File "/home/travis/build/qutech/qc-toolkit/qctoolkit/utils/sympy.py", line 189, in evaluate_lambdified
    [{'ceiling': numpy_compatible_ceiling}, 'numpy'])
[...]

PR #309, opened 2018-07-13
Issues/302 sympy free symbols

terrorfisch:
Makes qc-toolkit compatible with sympy 1.2 by wrapping `free_symbols` to get the old behavior.


terrorfisch:
Closes #302

PR #311, opened 2018-07-13
Increased test coverage for serialization

### LINKED ISSUE #294: Increase serialization overhaul test coverage

lumip:
Closes #294

I looked into coveralls to see which lines/branches are not yet covered by tests.
In particular the case in Serializable.__init__() where a pulse might get registered only after the gc is invoked was not covered. I tried to do this (delete a registered pulse with del while explicitely disabling the automatic gc) but couldn't cause the desired behavior. Are we sure that the invocation of the gc is necessary at that point?
Also, there's the issue of the AnonymousSerializable. It is currently not neatly integrated into the Serializable class tree and should really be (or be an abstract class defining abstract methods).

### DIFF
--- a/qctoolkit/serialization.py
+++ b/qctoolkit/serialization.py
@@ -414,12 +414,12 @@ class Serializable(metaclass=SerializableMeta):
             registry = default_pulse_registry
 
         if self.identifier and registry is not None:
-            if self.identifier in registry:
+            if self.identifier in registry and isinstance(registry, weakref.WeakValueDictionary):
                 # trigger garbage collection in case the registered object isn't referenced anymore
                 gc.collect(2)
 
-                if self.identifier in registry:
-                    raise RuntimeError('Pulse with name already exists', self.identifier)
+            if self.identifier in registry:
+                raise RuntimeError('Pulse with name already exists', self.identifier)
 
             registry[self.identifier] = self
 
--- a/qctoolkit/serialization.py
+++ b/qctoolkit/serialization.py
@@ -326,17 +326,6 @@ class DictBackend(StorageBackend):
         return set(self._cache.keys())
 
 
-def get_type_identifier(obj: Any) -> str:
-    """Return a unique type identifier for any object.
-
-    Args:
-        obj: The object for which to obtain a type identifier.
-    Returns:
-        The type identifier as a string.
-    """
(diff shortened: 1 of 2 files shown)


terrorfisch:
> Are we sure that the invocation of the gc is necessary at that point?
I think it can be removed as the default_storage_backend isn't a `WeakValueDictionary` by default anymore.

> Also, there's the issue of the AnonymousSerializable. It is currently not neatly integrated into the Serializable class tree and should really be (or be an abstract class defining abstract methods).
An ABCMeta metaclass makes calling isinstance much more costly(~10) and we call isinstance with `Expression` as type quite a lot. We could drop this and just check if the object we want to serialize has a `get_serialization_data` attribute here: https://github.com/qutech/qc-toolkit/blob/9548742bb565ce50171eacfd8e47eff02df6b0d8/qctoolkit/serialization.py#L861

terrorfisch:
I googled a bit. We should use `__getnewargs__` for this. I will update this pull request

lumip:
> I think it can be removed as the default_storage_backend isn't a WeakValueDictionary by default anymore.

hm.. but in principle it's still possible to set the default_storage_backend as a `WeakValueDictionary` so if the gc invocation is necessary in that case we shouldn't just remove that (although the tests I did worked fine without the gc even for WeakValueDictionary but that might be luck?)... However, the `weakref` documentation states that 

> if the referent is no longer alive, calling the reference object will cause None to be returned

so I guess, instead of invoking the garbace collector, we can just check whether the object currently stored in the registry (if any) is `None` ..?

> An ABCMeta metaclass makes calling isinstance much more costly(~10) and we call isinstance with Expression as type quite a lot. We could drop this and just check if the object we want to serialize has a get_serialization_data attribute here: 

-.-
well.. in that case I guess it's best to leave it as it is right now (and maybe try to find a nicer solution for the Expressions in the long run..?)

terrorfisch:
> hm.. but in principle it's still possible to set the default_storage_backend as a WeakValueDictionary so if the gc invocation is necessary in that case we shouldn't just remove that (although the tests I did worked fine without the gc even for WeakValueDictionary but that might be luck?)... However, the weakref documentation states that

IMHO, if the user sets it as an WeakValueDict, he is responsible for cleaning up. If desired we could provide tools for this.

terrorfisch:
Merge if you think this is ok.

Issue #149, opened 2018-01-18
Compensation for distortion and crosstalk


terrorfisch:
The user wants to define a pulse that is preferrably independent of experiment specific distortions(filters, bias-t, ...).

We need a way to make the compensation for that easy and general.

terrorfisch:
Furthermore one may wants to compensate the pulse in channel A by some modification to channel B and vice versa.

Currently I am preferring to put this into the ```HardwareSetup``` class. Currently there already is a mapping with an arbitary function. This needs to be extended to take all channels as arguments. This solves the crosstalk compensation issue.

The problem of predistortion can only be solved if the complete program is run through some kind of filter. It can obviously not be done locally as the transients between different waveforms have to be taken into account.

terrorfisch:
Together with @pcerf I came up with the idea to do this during the sequencing controlled by flags bound to an identifier. In this way it can be made pulse template specific.

lumip:
In the original design I would have placed it in a step after sequencing which would then carry out final logical channel to hardware mapping and also take care of hardware/setup specific error correction. However, I am not yet sufficiently aware of all the changes that were made to sequencing and waveforms to make a call here.
Doing the correction during the sequencing would mess up the clean separation of sequencing into an abstract waveform sequence from actual hardware requirements and make the whole process more complicated, so I'd refrain from doing that if it all possible and practical.

terrorfisch:
Implementation fo linear cross talk compensation in the `Waveform` class via a generalized `Transformation` class. There is no interface yet
https://github.com/qutech/qc-toolkit/tree/issues/149_linear_crosstalk_compensation

Based on #312

PR #316, opened 2018-07-19
Introduce waveform type that implements linear compensation of crosstalk

terrorfisch:
- Abstract `Transformation` class with a hopfully general enough interface using pandas
- Linear transformation implementation of said class
- `TransformingWaveform` that applies such a transformation
- `SubsetWaveform` which selects a subset of a `Waveform`s channels. This is needed as `TransformingWaveform` always calculates all channels.

TODO: Add pandas dependency to setup.py

Partially implements #149

### DIFF
--- a/qctoolkit/_program/transformation.py
+++ b/qctoolkit/_program/transformation.py
@@ -1,0 +1,55 @@
+from typing import Mapping, Set
+from abc import abstractmethod
+
+import numpy as np
+import pandas as pd
+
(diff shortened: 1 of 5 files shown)


terrorfisch:
Adds pandas to requirements. A very useful library for organizing, analyzing and manipulating tabular data.

If we want to shoot for fewer dependencies I can implement the "named rows and columns" feature we take from pandas myself but I would prefer sticking to the tested pandas code.

PR #317, opened 2018-07-19
Some refactoring

lumip:


qctoolkit/pulses/pulse_template_parameter_mapping.py
```diff
@@ -1,369 +1,8 @@
+from qctoolkit.pulses.mapping_pulse_template import MappingPulseTemplate
 
-from typing import Optional, Set, Dict, Union, List, Any, Tuple
-import itertools
-import numbers
+__all__ = [MappingPulseTemplate]
```
terrorfisch: `__all__` should be a list of strings.

https://docs.python.org/3/reference/simple_stmts.html#the-import-statement

PR #320, opened 2018-07-19
Added pulse from new serialization routines to backward compatitiblity tests

lumip:
These tests now test deserialization for both the old and the new serialization routines.


terrorfisch:
We should rename _loop to loop. It was not intended by me to have a private module in a private package.

Issue #287, opened 2018-07-05
Implementation of PT.create_program()


lumip:
See #286

lumip:
Now implemented for all PTs except MappingPT. @terrorfisch you said you had something for this already, aye?

PR #327, opened 2018-08-02
Changes to (_internal_)create_program()

lumip:
Implementations of new signature PT._internal_create_program for RepetitionPT, SequencePT, ForLoopPT, AtomicPT (and thus all subclasses of that) + tests.

MappingPT is the only one missing now.
After that is done, #287 should be finished.

### DIFF
--- a/qctoolkit/pulses/loop_pulse_template.py
+++ b/qctoolkit/pulses/loop_pulse_template.py
@@ -12,7 +12,7 @@ from qctoolkit._program._loop import Loop
 
 from qctoolkit.expressions import ExpressionScalar
 from qctoolkit.utils import checked_int_cast
-from qctoolkit.pulses.parameters import Parameter, ConstantParameter, InvalidParameterNameException, ParameterConstrainer
+from qctoolkit.pulses.parameters import Parameter, ConstantParameter, InvalidParameterNameException, ParameterConstrainer, ParameterNotProvidedException
 from qctoolkit.pulses.pulse_template import PulseTemplate, ChannelID
 from qctoolkit.pulses.conditions import Condition, ConditionMissingException
 from qctoolkit._program.instructions import InstructionBlock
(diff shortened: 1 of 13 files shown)


qctoolkit/pulses/pulse_template.py
```diff
@@ -118,7 +118,7 @@ def create_program(self, *,
         if parameters is None:
             parameters = dict()
         if measurement_mapping is None:
-            measurement_mapping = dict()
+            measurement_mapping = dict() # todo (2018-08-02): should not be empty but an identity mapping; requires PT to be a MeasurementDefiner (?)
```
terrorfisch: `measurement_names` is an abstract property of `PulseTemplate`

qctoolkit/pulses/pulse_template.py
```diff
@@ -190,19 +197,25 @@ def build_sequence(self,
...
-                      for parameter_name, parameter_value in parameters.items()
-                      if parameter_name in self.parameter_names}
+        try:
+            parameters = {parameter_name: parameters[parameter_name].get_value()
+                          for parameter_name in self.parameter_names}
+
+            measurement_parameters = {parameter_name: parameters[parameter_name].get_value()
+                                      for parameter_name in self.measurement_parameters}
+        except KeyError as e:
+            raise ParameterNotProvidedException(str(e)) from e
```
terrorfisch: Why do you create an exta dictionary for the measurement parameters? They should be a subset of `parameter_names`

qctoolkit/pulses/loop_pulse_template.py
```diff
@@ -228,28 +228,30 @@ def build_sequence(self,
...
-                                  for parameter_name in self.measurement_parameters}
+        try:
+            measurement_parameters = {parameter_name: parameters[parameter_name].get_value()
+                                      for parameter_name in self.measurement_parameters}
+            duration_parameters = {parameter_name: parameters[parameter_name].get_value()
+                                      for parameter_name in self.duration.variables}
+        except KeyError as e:
+            raise ParameterNotProvidedException(str(e)) from e
         measurements = self.get_measurement_windows(measurement_parameters, measurement_mapping)
+        if self.duration.evaluate_numeric(**duration_parameters) > 0:
```
terrorfisch: There should be a better way. But this is an optimization for the future.

qctoolkit/pulses/pulse_template.py
lumip: should be but currently aren't according to https://github.com/qutech/qc-toolkit/issues/230

qctoolkit/pulses/pulse_template.py
lumip: uh... done :D

PR #337, opened 2018-08-08
Compatibility Tests for create_program

lumip:
Extended tests for _internal_create_program to check that the result is the same as doing old Sequencing and constructing MultiChannelProgram from the resulting InstructionBlock.

Fixed behavior of AtomicPulseTemplate (and DummyPulseTemplate). Added measurements to compare_key of Loop.

Differences in behavior as detected by those tests are the following:

AtomicPulseTemplate (and thus all subclasses (TablePT, PointPT, MutliChannelPT, FunctionPT)):
- behavior exactly the same

SequencePulseTemplate:
- if all subtemplates have no duration (i.e. duration of the SequencePT is 0), _internal_create_program adds nothing (no children, no waveform, no measurements) to the parent_loop. Sequencer+MCP results in empty loop with measurements from SequencePT.

RepetitionPulseTemplate:
- if repetition count is zero or negative and a measurement was defined in RepetitionPT: Sequencer+MCP return an empty Loop with measurement, _internal_create_program does not modify the given parent_loop (does not add measurement)
- if the body template has duration zero: Sequencer+MCP return an empty loop with repetition count as specified by the RepetitionPT, _interal_create_program does not modify the given parent_loop (repetition count unchanged)
- if the body template has duration zero and a measurement was defined in RepetitionPT: Sequencer+MCP return an empty loop with repetition count as specified by the RepetitionPT and measurements, _interal_create_program does not modify the given parent_loop (repetition count unchanged and no measurements)
- if the RepitionPT does not have siblings in the pulse template tree and the subtemplate specifies only a waveform: Sequencer+MCP return a single Loop object holding the waveform with the repetition count specified by RepetitionPT, _internal_create_program creates a nested Loop structure (as in all other cases) [optimization in MCP in lines 347-360]

MappingPulseTemplate:
- behavior exactly the same

ForLoopPulseTemplate:
- if the body template has duration zero and a measurement was defined in ForLoopPT: Sequencer+MCP return an empty loop with measurements, _internal_create_program does not modify the parent_loop object (does not add measurements)

Differences in dealing with measurements were expected.

The optimization for RepetitionPT should not be done in _internal_create_program but during a subsequent optimization/hardware adaption step imo.

Setting the repetition count for empty loops should be a non-issue as well.

Your thoughts, @terrorfisch ?

### DIFF
--- a/qctoolkit/_program/_loop.py
+++ b/qctoolkit/_program/_loop.py
@@ -44,7 +44,7 @@ class Loop(Comparable, Node):
 
     @property
     def compare_key(self) -> Tuple:
-        return self._waveform, self.repetition_count, tuple(c.compare_key for c in self)
+        return self._waveform, self.repetition_count, self._measurements, tuple(c.compare_key for c in self)
 
     def append_child(self, loop: Optional['Loop']=None, **kwargs) -> None:
         # do not invalidate but update cached duration
--- a/qctoolkit/pulses/pulse_template.py
+++ b/qctoolkit/pulses/pulse_template.py
@@ -197,6 +197,10 @@ class AtomicPulseTemplate(PulseTemplate, MeasurementDefiner):
                                  measurement_mapping: Dict[str, Optional[str]],
                                  channel_mapping: Dict[ChannelID, Optional[ChannelID]],
                                  parent_loop: Loop) -> None:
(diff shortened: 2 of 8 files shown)


terrorfisch:
Looks good

PR #340, opened 2018-08-13
First draft for transformation/compensation integration in create_program

terrorfisch:
As a base for discussion.


terrorfisch:
Idea:

- `create_program`: public function
- `_create_program`: private function. Handles transformations and translation of subprograms to single waveforms. Calls `self._internal_create_program`.
- `_internal_create_program`: Pulse template specific program creation

terrorfisch:
I think as a first step I will stick to a global transformation that is applied to each pulse.

terrorfisch:
Dropped for now in favor of #341

PR #365, opened 2018-08-30
Fixed a bug in SequenceWaveform.unsafe_sample() which could cause a crash

qutech-lab:
For all sub-waveform except the first one the sample_times numpy array had dtype=object instead of dtype=float64 which led to a crash applying lambdas resulting from sympy.lambdify (e.g. when sampling FunctionWaveforms).

Added a specific test case to ensure the problem does not occur again. Also added the whole initial sample code that first encountered the bug in a new test file tests/pulses/bug_tests.py now intended to collect all reported bugs.

-- lumip

### DIFF
--- a/tests/pulses/bug_tests.py
+++ b/tests/pulses/bug_tests.py
@@ -1,0 +1,49 @@
+import unittest
+
+from qctoolkit.pulses.table_pulse_template import TablePulseTemplate
+from qctoolkit.pulses.function_pulse_template import FunctionPulseTemplate
+from qctoolkit.pulses.sequence_pulse_template import SequencePulseTemplate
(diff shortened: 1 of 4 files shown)


tests/pulses/bug_tests.py
```diff
@@ -0,0 +1,49 @@
...
+                                                                               ('n', 'tb', 'tend-tb')])
+
+        parameters = {'ta': 2,
+                      'va': 2,
+                      'tb': 4,
+                      'vb': 3,
+                      'tc': 5,
+                      'td': 11,
+                      'tend': 6}
+        _ = plot(template, parameters, sample_rate=100, show=False, plot_measurements={'m', 'n'})
```
terrorfisch: Doesnt the bug happen during rendering?
We cannot use plot without emulating a display in Travis CI.

PR #394, opened 2018-09-28
Added requirements.txt + some additional minor changes to testing

lumip:


qupulse/utils/types.py
```diff
@@ -5,11 +5,6 @@
 import fractions
 import warnings
 
-try:
```
terrorfisch: Are you shure this caused the error and not something nondeterministic?

terrorfisch: I still had errors locally. Should be solved with https://github.com/qutech/qupulse/pull/394/commits/f4982a7d07b8d02651ea708bf181250be0c5aad9

PR #400, opened 2018-10-02
[DO NOT MERGE] Parameter Namespaces [Variant A]

lumip:
Functionality to put parameters in namespaces (in a syntactically nice way); 
monkey patched sympy's auto_symbol token transform parsing step to allow symbol names with dots (e.g. "foo.bar", "foo.bar.bar") to signalize namespaces for parameters (instead of python object attribute access).

internally these will be replaced with underscore namespace notation `foo____bar____bar` (python prevents us from keeping the dots: symbol names are regularly python `eval`'d or used as keyword arguments in lamdification -> Syntax errors)

evaluation and substitution code in utils.sympy transparently renames incoming parameter dictionaries from dot to underscore notation (and back) if required. users only need to face dot notation.

**compared to Variant B, this has the following advantages:**
- arguably nicer notation `foo.bar.bla` instead of `NS(foo).NS(bar).bla`

**disadvantages:**
- internal mapping from dot to quadruple-underscore for namespace separation might be confusing during debugging internas
- using dots for namespaces precludes usage of dots for attribute access (which is generally possible in sympy expression but probably not really a feature qupulse needs to support)

**we currently favor Variant B**, any active development will be happening there


lumip:
I also copied the patched auto_symbol method into a sympy build and ran all sympy tests to guarantee that it doesn't break existing behavior. Currently, there are a few errors which need to be addressed.

lumip:
> I also copied the patched auto_symbol method into a sympy build and ran all sympy tests to guarantee that it doesn't break existing behavior. Currently, there are a few errors which need to be addressed.

One of the errors comes from a test which evaluated
```
x = Symbol('x')
assert parse_expr('Symbol("x").free_symbols') == x.free_symbols
````
showing clearly that the dot '.' can be (and IS) used in sympy expressions to access object attributes. This basically makes using the dot as a separator for parameter namespaces impossible for us (if we want to support full sympy compliance, which would be good)..
Can we think of a better separator symbol?

terrorfisch:
A discussion with @pcerf @reneotten and others resulted in choosing Variant B ( #418 ).

Issue #402, opened 2018-10-05
Allow check of AtomicMultiChannelPulseTemplate subtemplate duration on instantiation


terrorfisch:
Currently, it is required that the durations of AtomicMCPT's subtemplates are all equal on creation.

This can be tedious. We should change this into an instantiation requirement.

Preferably by a new keyword argument `duration: [Expression, bool]=False`.
```python
def __init__(..., duration: [Expression, bool]=False):
    # ...
    if not duration:
        # old checks of equal duration
       self._duration = None
    elif duration is True:
        self._duration = subtemplates[0].duration
    else:
        self._duration = Expression(duration)

def build_waveform(...):
    # ...
    if self._duration:
        # assert all subwaveform durations are equal self._duration.evaluate_numeric(**parameters)
```

PR #403, opened 2018-10-08
Issues/402 multi channel pt duration

### LINKED ISSUE #401: duration calulation comparision fails for combined pulses
When combining two templates in `AtomicMultiChannelPulseTemplate.__init__` I get the following error message:
```
ValueError: Could not assert duration equality of 11100647.5000000 and 11100647.5000000
```
The first template I created directly, the other is a concatenation of multiple templates.

The numbers in the error message appear equal, so I investigated.
```
ipdb> duration
Expression('11100647.5000000')

ipdb> subtemplate.duration
Expression('11100647.5000000')

ipdb> type(duration)
<class 'qupulse.expressions.ExpressionScalar'>

ipdb> float(duration)==float(subtemplate.duration)
True

ipdb> duration._sympified_expression._mpf_
(0, mpz(5959614747115521), -29, 53)

ipdb> subtemplate.duration._sympified_expression._mpf_
(0, mpz(22201295), -1, 25)
```
The comparison that fails is this one: https://github.com/qutech/qupulse/blob/master/qupulse/expressions.py#L269

It looks like the equality check for an `ExpressionScalar` fails for number which are _almost_ identical, except for some finite precision rounding errors. For the comparison of the `duration` of pulses I think the comparison should be less strict.

@terrorfisch

terrorfisch:
Implement #402 
Fix #401

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,6 @@ PyVISA==1.9.1
 sympy==1.3
 gmpy2==2.0.8
 
-certifi==2018.8.24
 cycler==0.10.0
 kiwisolver==1.0.1
 mpmath==1.0.0
--- a/ReleaseNotes.txt
+++ b/ReleaseNotes.txt
@@ -1,5 +1,13 @@
 ## pending/current ##
 
+- General:
+    - Introduce qupulse.utils.isclose (an alias for math.isclose if available)
+
+- Pulse Templates:
+    - `AtomicMultichannelPulseTemplate`:
+        - Add duration keyword argument & example (see MultiChannelTemplates notebook)
+        - Make duration equality check approximate (numeric tolerance)
+
 - Expressions:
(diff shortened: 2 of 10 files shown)


requirements.txt
```diff
@@ -5,7 +5,6 @@ PyVISA==1.9.1
 sympy==1.3
 gmpy2==2.0.8
 
-certifi==2018.8.24
```
terrorfisch: @lumip 
certifi cannot be downgraded as it is a distutils package. Why do we fix its version?
https://travis-ci.org/qutech/qupulse/jobs/442140464

terrorfisch:
The tests pass using python=3.4 although we use math.isclose which is introduced in python 3.5
https://docs.python.org/3.4/library/math.html

A local python=3.4 install from conda fails.

PR #415, opened 2018-11-08
Fix/numeric expression evaluation

terrorfisch:
TODO: Determine the reason for these changes and create tests.


terrorfisch:
No changes -> delete branch

PR #418, opened 2018-11-28
[WIP] Parameter Namespaces [Variant B]

### LINKED ISSUE #344: Introduce Namespaces for Parameters
It should be able to put parameters in a namespace, i.e., a group of parameters whose values depend on some specific context, e.g., on the specific qubit they are executed on.

lumip:
Functionality to put parameters in namespaces, alternative to #400
added specialized classes `NamespacedSymbol`, `NamespacedIndexedBase` and `SymbolNamespace`. Adapated routines to identify subscripted symbols. Allows symbol names with explicit namespacing, e.g. "NS(foo).bar", "NS(foo).NS(bar).bar" to signalize namespaces for parameters. These symbols will be represented by the classes above.

**compared to Variant A, this is a more explicit namespace notation with two advantages:**
- no internal renaming of namespaced parameters -> nicer/less confusing debugging (renaming still takes place in evaluate_lamdified but nobody ever sees that)
- due to separate SymbolNamespace class, no conflict with accesses to symbol attributes

todo:
- ensure that namespaced parameters work with all PulseTempaltes (TablePT and MappingPT done)

note:
branch `issues/344_parameter_namespaces_3_attempt_at_no_renaming_for_lamdified_val` is an attempt to forego flattening of nested parameter dictionaries by printing lamdified code that directly works on nested dicts. however, it currently does not work

when merged, closes #344


shumpohl:
Sadly moved out of scope

PR #426, opened 2019-02-26
[WIP] First draft for virtual AWG

terrorfisch:
Not final.

Expand by a more qutip friendly interface that provides a sampling function.


terrorfisch:
I decided to drop the fixed sample rate interface for now as there is no immediate usecase. Instead I implemented a function handle callback. This is the interface needed for qutip integration.

Issue #422, opened 2019-02-21
Possibillity to add (constant valued) channel to non-atomic pulse template


terrorfisch:
Example use case:

The user wrote a complex pulse template. The user now wants to add a channel with a constant value to that pulse template (for example a marker channel with value '0').

Idea:
Wrap all returned waveforms with a `MultiChannelWaveform` that sets the added channels to the given value.

terrorfisch:
Was implemented with #429

PR #429, opened 2019-02-26
Issues/422 parallel constant channel

terrorfisch:
Implement #422 

Includes example: yay!
Tests incomplete: nay :(


terrorfisch:
Added tests

PR #434, opened 2019-03-11
Issues/433 plotting isinstance fail

### LINKED ISSUE #433: `plotting.render()` returns `None` if isinstance checks fail.
This leads to a `TypeError` when `plot()` calls `render()`.

Maybe related? #261

terrorfisch:
Fixes #433 which occures if a reload is triggered (ipython and spyder do this sometimes).

I implemented a custom `has_type_interface`. Something tells me that there has to be a better more standard way.

### DIFF
--- a/qupulse/pulses/plotting.py
+++ b/qupulse/pulses/plotting.py
@@ -139,7 +139,7 @@ def render(program: Union[AbstractInstructionBlock, Loop],
             raise ValueError("Keyword argument time_slice is not supported when rendering instruction blocks")
         return _render_instruction_block(program, sample_rate=sample_rate, render_measurements=render_measurements)
     else:
-        raise ValueError('Cannot render an object of type %s', type(program))
+        raise ValueError('Cannot render an object of type %r' % type(program), program)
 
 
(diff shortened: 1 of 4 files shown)


qupulse/pulses/plotting.py
```diff
@@ -130,15 +130,16 @@ def render(program: Union[AbstractInstructionBlock, Loop],
...
+    elif has_type_interface(program, AbstractInstructionBlock):
         warnings.warn("InstructionBlock API is deprecated", DeprecationWarning)
         if time_slice is not None:
             raise ValueError("Keyword argument time_slice is not supported when rendering instruction blocks")
         return _render_instruction_block(program, sample_rate=sample_rate, render_measurements=render_measurements)
-    elif isinstance(program, Loop):
-        return _render_loop(program, sample_rate=sample_rate,
-                            render_measurements=render_measurements, time_slice=time_slice)
+    else:
+        raise ValueError('Cannot render an object of type %s', type(program))
```
terrorfisch: This is bad. There should be a `%`

terrorfisch:
Apparently the mocking of sys.modules does not work.

PR #448, opened 2019-04-03
Fix #447

### LINKED ISSUE #447: Exception when rendering TablePT
`qupulse` generated an exception when trying to render a simple pulse with repetitions. The exception seems to be caused by rounding errors. A minimal example:

``` python
#%% Exception due to rounding error in qupulse
# qupulse.__version__ 0.3
# gmpy2 installed
from qupulse.pulses import TablePT
from qupulse.pulses.sequencing import Sequencer as Sequencing
from qupulse.pulses import SequencePT
from qupulse.pulses.plotting import render

period=8.192004194306148e-05
repetitions=80
sampling_rate=1e7
sec_to_ns=1e9

table_pt= TablePT({'test': [(0, 0), (period * sec_to_ns, 0, 'linear')]})
    
sequencer = Sequencing()
template = SequencePT(*((table_pt,) * repetitions))
channels = template.defined_channels
sequencer.push(template, dict(), channel_mapping={ch: ch for ch in channels},
               window_mapping={w: w for w in template.measurement_names})
instructions = sequencer.build()
(_, voltages, _) = render(instructions, sampling_rate / sec_to_ns)
```

This generated the following error:
```
...
  File "c:\projects\qupulse\qupulse\pulses\plotting.py", line 176, in _render_instruction_block
    output_array=output_array)

  File "c:\projects\qupulse\qupulse\_program\waveforms.py", line 84, in get_sampled
    raise ValueError('The sample times are not in the range [0, duration]')

ValueError: The sample times are not in the range [0, duration]
```
When setting the number of repetitions lower (e.g. 8), the exception does not occur.

Might be related to #176

@terrorfisch

terrorfisch:
Fix #447 by avoiding duplication and cheating a bit. The check which throws the error is not done in `SequenceWaveform.get_sample`.

@peendebak

### DIFF
--- a/ReleaseNotes.txt
+++ b/ReleaseNotes.txt
@@ -7,6 +7,8 @@
 - Pulse Templates:
     - `MappingPulseTemplate`:
         - Raise a ValueError if more than one inner channel is mapped to the same outer channel
+    - Plotting:
+        - Make `plotting.render` behaviour and return value consistent between calls with `InstructionBlock` and `Loop`. Render now always returns 3 arguments.
 
 ## 0.3 ##
 
--- a/qupulse/pulses/plotting.py
+++ b/qupulse/pulses/plotting.py
@@ -14,10 +14,10 @@ import warnings
 import operator
 import itertools
 
-from qupulse.utils.types import ChannelID, MeasurementWindow, has_type_interface
+from qupulse.utils.types import ChannelID, MeasurementWindow, has_type_interface, TimeType
 from qupulse.pulses.pulse_template import PulseTemplate
 from qupulse.pulses.parameters import Parameter
-from qupulse._program.waveforms import Waveform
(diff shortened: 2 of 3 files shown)


peendebak:
@terrorfisch The PR solves the issue #447 I reported. My only comments is that in the example in #447 the `voltages['test']` is a read-only array. I can easily fix this in my application, but it means the PR is not completely backwards compatible.

terrorfisch:
@peendebak Good catch. This was an inconsistency between `_render_instruction_block` and `_render_loop` before. I think making both writable is the way to go as the result caching is not necessary here anyways. Could you check whether https://github.com/qutech/qupulse/pull/448/commits/5b5b5aae5363078fea8bebb8a3c6980d096b7a0f still breaks your code or produces any other inconsistent/undesired behaviour on your side?

peendebak:
@terrorfisch Looks good

PR #457, opened 2019-05-09
QCoDeS integration

qutech-lab:
This is needed for QCoDeS to determine the shape of some operation results


terrorfisch:
With tests in  #459

Issue #446, opened 2019-04-01
Make offset and amplitude handling of AWG configurable


terrorfisch:
Currently the AWG-drivers treat the incoming waveforms as if their unit is Voltage and take the respective channels output range into account. However, the offset is ignored. This is because it turned out to be convenient to quickly sweep the offset of a pulse via the offset setting of the awg.

The downside of this is, that the user might expect that the AWG's offset is taken into account when uploading the waveform. I think it is reasonable to add the functionality of ignoring this offset to the AWG interface.

Objections:
 - Some AWGs can change such a setting themselves at runtime (HDAWG)

PR #463, opened 2019-05-23
Tektronix AWG with offset handling

qutech-lab:
This will pull a version of the Tektronix driver which implements offset handling as described in #446


terrorfisch:
I want to merge this asap. Blocking: TekAwg is not on PyPi.

Two options:
1. Cleanup TekAwg a bit (and rename it to tek_awg) to upload it to pypi
2. Migrate to qcodes instrument driver as a "backend"

I think the second would be more work for now but makes sense in the long run.

terrorfisch:
See #464

PR #466, opened 2019-06-09
allow ExpressionScalar to be used in f-strings

eendebakpt:
We might have to discuss what is the desired behaviour for f-strings, but it seems reasonable to allow printing as a floating point number.

We also need to disable tests for python 3.5. Any preferred way to do this?

@terrorfisch

### DIFF
--- a/qupulse/expressions.py
+++ b/qupulse/expressions.py
@@ -240,6 +240,11 @@ class ExpressionScalar(Expression):
     def __repr__(self) -> str:
         return 'Expression({})'.format(repr(self._original_expression))
 
+    def __format__(self, format_spec):
+        if format_spec=='':
+            return str(self)
+        return f'%{format_spec}' % float(self)
+    
     @property
     def variables(self) -> Sequence[str]:
         return self._variables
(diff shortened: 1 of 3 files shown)


qupulse/expressions.py
```diff
@@ -240,6 +240,11 @@ def __str__(self) -> str:
     def __repr__(self) -> str:
         return 'Expression({})'.format(repr(self._original_expression))
 
+    def __format__(self, format_spec):
+        if format_spec=='':
+            return str(self)
+        return f'%{format_spec}' % float(self)
```
terrorfisch: Can't we just replace this with
```python
return format(float(self), format_spec)
```

terrorfisch:
I moved the fstring test into an indirect `exec` of source code to allow parsing the test file with python 3.5

qupulse/expressions.py
eendebakpt: Seems good, I didn't know about the `format` method.

eendebakpt:
@terrorfisch Thanks for updating the good. Looks good to me

PR #472, opened 2019-07-30
allow MappingPT to remove channels

eendebakpt:


terrorfisch:
I would rather implement it as an explicit mapping to `None` instead of an implicit drop. Otherwise the user might expect that not mapping a channel keeps it the same.

terrorfisch:
Maybe we should introduce some kind of helper function like `drop_unmapped` and `forward_unmapped` to make constructing mappings more convenient.

```python
def drop_unmapped_channels(inner_template, mapping):
    return {inner_channel: mapping.get(inner_channel, None) for inner_channel in inner_template.defined_channels}

def forward_unmapped_channels(inner_template, mapping):
    return {inner_channel: mapping.get(inner_channel, inner_channel) for inner_channel in inner_template.defined_channels}
```

PR #489, opened 2020-01-07
Change loop repetition at runtime

terrorfisch:
TODO:
 - [x] Tests for `create_program` implementations
 - [ ] Tests for `flatten_and_balance`, `make_compatible` and other Loop modification functions
 - [x] Warn if a parameter constraint depends on a volatile parameter
 - [x] Implement functionality for TaborAWG (@Hylta )
 - [x] Implement functionality for Zurich Instruments

Problems:
 - HDAWG driver crashes python when using multiple awgs via `HardwareSetup`


terrorfisch:
Also fixes #492

PR #500, opened 2020-01-17
Feat/driver abstraction

bpapajewski:
Implemented abstract AWG Drivers. Test cases and examples will follow soon.

 - [ ] Remove `old_base.py`


qupulse/hardware/awgs/base.py
```diff
@@ -1,166 +1,202 @@
...
+        for attr in directory:
+            if callable(getattr(type(self), attr)) and attr[0] != "_":
+                if not (attr in function_list):
+                    function_list[attr] = getattr(self, attr)
+        return function_list
 
-    _valid = [IGNORE_OFFSET, CONSIDER_OFFSET]
+    @property
+    def functions(self) -> Dict[str, Callable]:
+        """Returns a copy of the dictionary with all public functions of the feature"""
```
terrorfisch: Probably better to return a MappingProxyType

qupulse/hardware/awgs/base.py
```diff
@@ -1,166 +1,202 @@
...
+        Reads the functions of a feature and returns them as a dictionary
 
-class AWGAmplitudeOffsetHandling:
-    IGNORE_OFFSET = 'ignore_offset'   # Offset is ignored.
-    CONSIDER_OFFSET = 'consider_offset' # Offset is discounted from the waveforms.
-    # TODO OPTIMIZED = 'optimized' # Offset and amplitude are set depending on the waveforms to maximize the waveforms resolution
+        Return:
+            Returns dictionary with all functions of the feature
+        """
+        directory = dir(self)
```
terrorfisch:
```python
return {name: attr
        for name, attr in inspect.getmembers(self, inspect.ismethod)
        if not name.startswith('_')}
```

qupulse/hardware/awgs/base.py
```diff
@@ -1,166 +1,202 @@
...
-Classes:
-    - AWG: Common AWG interface.
-    - DummyAWG: A software stub implementation of the AWG interface.
-    - ProgramOverwriteException
-    - OutOfWaveformMemoryException
-"""
 
-from abc import abstractmethod
-from typing import Set, Tuple, Callable, Optional
+class BaseFeature(ABC):
```
terrorfisch: Does it make sense to make BaseFeature a metaclass for better static analysis?

lankes-fzj: What do you mean with "static analysis", @terrorfisch ? Do you mean the syntax checking, that is performed by the IDE (or mypy, for example)? If yes, then I didn't find a way that really works. If you have an idea to solve this, feel free to show us :)

lankes-fzj:
I've added a test script for the base drivers in 5e1c6216238bb1f0feff3b557e1727b7cab77be4. This is not a unit test, yet. Firstly, it just serves as an example of how to use the drivers

qupulse/hardware/awgs/base.py
terrorfisch: Yes, basically for IDE support. It works with warnings for abstractmethods but maybe because this is hard baked into the analyzers. I'll test my idea with jedi which is the analyzer Spyder uses afaik.

terrorfisch:
What do you think about implementing features like this
```python
class MyAmplitudeFeature(AmplitudeFeature):
    """specific to MyAWG"""
    def get_amplitude(self) -> float:
        return 42

my_awg = MyAWG()

my_awg[AmplitudeFeature].get_amplitude()
# or
my_awg.get_feat(AmplitudeFeature).get_amplitude()
```
Pro:
 - I think it should be possible to implement FeatureAble's `__getitem__` or `get_feature` static analiyzer friendly. Currently I like the `__getitem__` syntax more.
 - This avoids name collisions between features

Cons:
 - Noisier usage

lankes-fzj:
> What do you think about implementing features like this
> ...

I also had a similar idea to implement the features. The problem was that you also have to pick the right feature object from a list. My goal was to allow users to call the feature like a normal function of the driver object.

But if you, as a future user, prefer a solution like this, we can do it. The question is what solution do you find more comfortable?

terrorfisch:
> Additionally, the specific feature has no access to the driver's attributes.

What do you mean by that?

> The question is what solution do you find more comfortable?

I think for usage in the interpreter the proposed addition af attributes is more convenient. For developing I like the explicit feature access more. I think the main usage of the driver objects will be in other library code not by the user itself so I would opt for the explicit feature access. We could provice a utility class/method to create a "unifying object" on top if needed while this is very hard in the other direction. Something like:


```python
class Unifyer:
    def __init__(self, featureable):
        self._wrapped = featureable
    
    def __lookup_iter(self):
        yield self._wrapped
        yield from self._wrapped.features().values()
    
    def __getattr__(self, attr):
        for obj in self.__lookup_iter():
            try:
                return getattr(feature, attr)
            except AttributeError:
                pass
        else:
            raise AttributeError(attr)
    
    def __dir__(self):
        return sum(dir(obj) for ob in self.__lookup_iter())
```

lankes-fzj:
> > Additionally, the specific feature has no access to the driver's attributes.
> 
> What do you mean by that?

What I meant, was that the feature object has no access to the hardware connections of the driver object because these are the driver's attributes. But I just realised, that it doesn't matter if we provide a callable to the feature or the driver object that allows the feature itself to contact the hardware.

So I would say, we will implement your idea, to make it clearer and more comfortable.

terrorfisch:
This works in pycharm but spyder i.e. jedi (14.1) struggles with it:

```python
import typing

class Feature:
    pass

class MyBase(Feature):
    def the_method(self):
        pass

class MyDerived(MyBase):
    def the_method(self):
        pass

T = typing.TypeVar('T', bound=Feature)

class MyThing:
    def __getitem__(self, feature_type: typing.Type[T]) -> T:
        pass
```

> So I would say, we will implement your idea, to make it clearer and more comfortable.

 I guess this is a good start.

lankes-fzj:
> * Is there a reason for prefixing some base class names with "Base"?

We can change this, if you want. These classes were named like this, so that everyone knows that these are abstract base classes.

> * What is the reasoning behind moving the old AWG class to `old_base.py`? I'd say either we keep it for legacy reasons then it should stay in `base.py` or we ditch it completely.

As soon as the drivers are translated into the new class hierarchy, the `old_base.py` is not needed anymore. I just wanted to keep the old drivers until they are available with the new syntax.

terrorfisch:
> We can change this, if you want. These classes were named like this, so that everyone knows that these are abstract base classes.

In my optinion it adds avoidable visual noise.

> As soon as the drivers are translated into the new class hierarchy, the `old_base.py` is not needed anymore. I just wanted to keep the old drivers until they are available with the new syntax.

I'll add a TODO item for removing.

What AWG do you target first?

lankes-fzj:
Benjamin is currently translating the Tabor driver into the new structure, but this happens in another branch.

lankes-fzj:
@terrorfisch, is it possible that the CI tool has problems with sympy?

terrorfisch:
> @terrorfisch, is it possible that the CI tool has problems with sympy?

Was fixed here: https://github.com/qutech/qupulse/pull/491

lankes-fzj:
@terrorfisch, I made some final changes and I think that the driver framework is ready now. Unless you have any remarks, we should finish this pull request so that we can focus on the specific awg drivers.

lankes-fzj:
@terrorfisch, I made some final changes and I think that the driver framework is ready now. Unless you have any remarks, we should finish this pull request so that we can focus on the specific awg drivers.

Unfortunately, the change list of this pull request became a bit confusing due to the merges I just did.

terrorfisch:
> Unfortunately, the change list of this pull request became a bit confusing due to the merges I just did.

I rebased your commits and made a new pull request here: #509

PR #502, opened 2020-01-20
improve warning for plotting of non-integer number of samples

peendebak:
The plotting method issues a warning of the number of samples to be plotted is non-integer. This PR prevents issuing the warning in case of rounding errors.

### DIFF
--- a/qupulse/pulses/plotting.py
+++ b/qupulse/pulses/plotting.py
@@ -167,8 +167,8 @@ def render(program: Union[AbstractInstructionBlock, Loop],
     if sample_count < 2:
         raise PlottingNotPossibleException(pulse=None,
                                            description='cannot render sequence with less than 2 data points')
-    if not float(sample_count).is_integer():
-        warnings.warn("Sample count is not an integer. Will be rounded (this changes the sample rate).")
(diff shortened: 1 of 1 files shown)


qupulse/pulses/plotting.py
```diff
@@ -167,8 +167,8 @@ def render(program: Union[AbstractInstructionBlock, Loop],
     if sample_count < 2:
         raise PlottingNotPossibleException(pulse=None,
                                            description='cannot render sequence with less than 2 data points')
-    if not float(sample_count).is_integer():
-        warnings.warn("Sample count is not an integer. Will be rounded (this changes the sample rate).")
+    if not round(float(sample_count), 10).is_integer():
+        warnings.warn(f"Sample count {sample_count} is not an integer. Will be rounded (this changes the sample rate).")
```
terrorfisch: Can you change this to
```python
warnings.warn("Sample count {sample_count} is not an integer. Will be rounded (this changes the sample rate).".format(sample_count=sample_count)
```

Unfortunately we still need to support python 3.5

terrorfisch:
Thanks for the contribution. Looks good apart from the f-string. I wish we could use that.

PR #503, opened 2020-01-20
Performance improvements with numba

terrorfisch:
Use numba for voltage -> uint16 binary conversion. This is a bottleneck for pulses that use many small waveforms (like charge scans).


terrorfisch:
@peendebak numba promises a lot of possible imprevements in binary waveform handling code but it requires llvm i.e. some configuration or a conda environment under windows. Would this be problematic for you?

PR #505, opened 2020-01-22
Added first TaborAWG support for issue/342

Hylta:
First draft of set_volatile_parameters function and changed TaborProgram class correspondingly


qupulse/hardware/awgs/tabor.py
```diff
@@ -239,20 +243,28 @@ def setup_single_sequence_mode(self) -> None:
 
         sequencer_table = []
         waveforms = OrderedDict()
+        position = -1
```
terrorfisch: Better use enumerate

qupulse/hardware/awgs/tabor.py
```diff
@@ -165,6 +166,9 @@ def __init__(self,
         self._waveforms = []  # type: List[MultiChannelWaveform]
         self._sequencer_tables = []
         self._advanced_sequencer_table = []
+        self._volatile_parameter_mappings = dict()  # keys: volatile parameter names, entries: position as list:
```
terrorfisch: Consider a named tuple for the entries. This makes access much easier.

qupulse/hardware/awgs/tabor.py
```diff
@@ -1167,6 +1201,70 @@ def remove(self, name: str) -> None:
         self.free_program(name)
         self.cleanup()
 
+    def set_volatile_parameters(self, program_name: str, parameters: Mapping[str, ConstantParameter]):
+        """Set the values of parameters which were marked as volatile on program creation."""
+        # TODO: Add exception if program creation threw a VolatileModificationWarning, add documentation
+
+        waveform_to_segment_index, program = self._known_programs[program_name]
+        names = [name for name in [*parameters] if name in [*program._volatile_parameter_mappings]]
```
terrorfisch:
```python
names = set(parameters.keys()) & set(program._volatile_parameter_mappings.keys())
```

qupulse/hardware/awgs/tabor.py
```diff
@@ -1167,6 +1201,70 @@ def remove(self, name: str) -> None:
...
+        if not names:
+            print('{} has no volatile parameters with these names'.format(program_name))
+            return
+
+        # Change sequencing tables in active program and program memory
+        if program.program.depth() == 1:
+            for name in names:
+                for program_mapping in program._volatile_parameter_mappings[name]:
+                    if program_mapping[0] == 0:
```
terrorfisch: If you make the entries named tuples this is much easier to read

qupulse/hardware/awgs/tabor.py
```diff
@@ -1167,6 +1201,70 @@ def remove(self, name: str) -> None:
...
+                        new_rep_count = int(program.program.children[program_mapping[2]].repetition_parameter.get_value())
+                        program.program.children[program_mapping[2]].repetition_count = new_rep_count
+                        if program_mapping[3] == 1:
+                            seg_ind = program._sequencer_tables[program_mapping[1]][program_mapping[2]][1]
+                            jump_flag = program._sequencer_tables[program_mapping[1]][program_mapping[2]][2]
+                            program._sequencer_tables[program_mapping[1]][program_mapping[2]] = (new_rep_count, seg_ind,
+                                                                                                 jump_flag)
+                            if self._current_program == program_name:
+                                self.device.send_cmd(':SEQ:SEL {}'.format(program_mapping[1] + 2))
+                                self.device.send_cmd(
```
terrorfisch: I would collect all changes to the current program in one place (a list or set of commands) before executing them. This makes testing a lot easier and the code possibly more performant.

qupulse/hardware/awgs/tabor.py
```diff
@@ -142,6 +142,8 @@ class TaborSequencing(Enum):
     ADVANCED = 2
 
 
+volatile_mapping_entry = namedtuple('Mapping_Entry', 'aSeq adv_position position change_table')
```
Hylta: Is it good style to use the namedtuple like this? Also are the names unterstandable?

terrorfisch: I would
1. keep class names CamelCase
2. The type name and the first argument to namedtuple should be the same
3. Better use typing.NamedTuple to annotate types
4. Keep member names snake_case

```python
from typing import NamedTuple

MappingEntry = NamedTuple('MappingEntry', [('a_seq', int), ('adv_position', int), ...]
```

qupulse/_program/_loop.py
```diff
@@ -653,6 +653,7 @@ def _is_compatible(program: Loop, min_len: int, quantum: int, sample_rate: TimeT
             if program.repetition_parameter is not None:
                 warnings.warn("_is_compatible requires an action which drops volatility.",
                               category=VolatileModificationWarning)
+                program._repetition_parameter = None
```
terrorfisch: We should not do this here. `_ic_compatible` should not modify the arguments

PR #509, opened 2020-02-05
Driver abstraction (rebased)

terrorfisch:


lankes-fzj:
@terrorfisch, can we merge this pull request and then use a new one for the specific drivers? Or do you want us to do everything in this pull request?

terrorfisch:
> @terrorfisch, can we merge this pull request and then use a new one for the specific drivers? Or do you want us to do everything in this pull request?

I plan to do a the 0.5 very soon and I do not think this feature will make it, so I think we should collect all changes and merge them at once.

PR #528, opened 2020-04-15
Feat/new tabor driver rebase

bpapajewski:
A Pullrequest for the tabor driver with the new driver abstraction


qupulse/hardware/awgs/tabor.py
```diff
@@ -1,67 +1,262 @@
...
 import teawg
-import numpy as np
 
-from qupulse.utils.types import ChannelID
-from qupulse._program._loop import Loop, make_compatible
-from qupulse.hardware.util import voltage_to_uint16, find_positions
-from qupulse.hardware.awgs.base import AWG, AWGAmplitudeOffsetHandling
-from qupulse._program.tabor import TaborSegment, TaborException, TaborProgram, PlottableProgram, TaborSequencing,\
-    make_combined_wave
+# What does this mean?
```
terrorfisch, line 29: That the driver assumes the native byte ordering is little endian

tests/hardware/awg_new_driver_base_tests.py
```diff
@@ -0,0 +1,270 @@
...
+import warnings
+
+from qupulse import ChannelID
+from qupulse._program._loop import Loop
+from qupulse.hardware.awgs_new_driver.base import AWGDevice, AWGChannel, AWGChannelTuple, AWGMarkerChannel
+from qupulse.hardware.awgs_new_driver.features import ChannelSynchronization, ProgramManagement, VoltageRange, \
+    AmplitudeOffsetHandling
+from qupulse.utils.types import Collection
+
+warnings.simplefilter("ignore", UserWarning)
```
terrorfisch, line 12: Please use a context manger to catch warnings when necessary. Otherwise this silences warnings we want occure during testing.

tests/hardware/tabor_tests.py
```diff
@@ -51,6 +49,4 @@ def test_config_guard(self):
         with self.assertRaises(RuntimeError):
             channel_pair.guarded_method(1, True)
 
-        self.assertFalse(channel_pair.is_in_config_mode)
-
-
+        self.assertFalse(channel_pair.is_in_config_mode)
```
terrorfisch, line 52: Please remove changes that do not change anything to make `git blame` more useful.

.gitignore
```diff
@@ -9,5 +9,7 @@ dist/*
 doc/source/examples/.ipynb_checkpoints/*
 **.asv
 *.orig
-MATLAB/+qc/personalPaths.mat
-/doc/source/_autosummary/*
+.idea/
+.mypy_cache/*
+.gitignore
+tests/hardware/WX2184C.exe
```
terrorfisch, line 15: Please do not remove `MATLAB/+qc/personalPaths.mat`.
Why do you ignore `.gitignore`?
Please add a newline at the end.

terrorfisch:
I think a good name for the sub-package would be `feature_awg`.

.gitignore
bpapajewski: I have ignored the gitignore with me, so that e.g. the WX2184C.exe does not stand with in the gitignore of everyone.
But since it is the gitignore in the git, ignoring it makes no sense anymore, so i removed it from the gitignore.
I also added MATLAB/+qc/personalPaths.mat and /doc/source/_autosummary/* again.

tests/hardware/awg_new_driver_base_tests.py
bpapajewski: I removed it

tests/hardware/tabor_tests.py
bpapajewski: Do you mean I should undo things like the changed formatting? 
And should I minimize merge conflicts in this file only?

terrorfisch: > Do you mean I should undo things like the changed formatting?

I this case yes. If we squash the commits before merging the file will not be touched.

> And should I minimize merge conflicts in this file only?

It is always a payoff.

bpapajewski: The unnecessary merge conflicts in tabor_tests.py should now be removed

bpapajewski:
> 
> 
> I think a good name for the sub-package would be `feature_awg`.

The sub-package is now renamed to `feature_awg`

tests/hardware/tabor_tests.py
bpapajewski: I added newlines to the end of the file, so it matches the one in the master branch.
But I do not think that the changes can simply remove.  In the first commit with this file, the spaces were already not there. I guess PyCharm already removed them automatically when adding the file.

terrorfisch:
See #557

PR #558, opened 2021-03-15
performance improvement for sampling

peendebak:
This PR improves sampling speed of the `get_sample_times` method.

```
%timeit np.arange(10000)/2.4
21 µs ± 132 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

%timeit np.arange(0, 10000/2.4, 1/2.4)
6.81 µs ± 28.2 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

a=np.arange(10000)/2.4

b=np.arange(0, 10000/2.4, 1/2.4)

a-b
Out[184]: array([0., 0., 0., ..., 0., 0., 0.])

np.max(np.abs(a-b))
Out[185]: 0.0
```

@terrorfisch

### DIFF
--- a/qupulse/hardware/util.py
+++ b/qupulse/hardware/util.py
@@ -95,7 +95,6 @@ def get_sample_times(waveforms: Union[Collection[Waveform], Waveform],
         segment_lengths.append(rounded_segment_length)
 
     segment_lengths = np.asarray(segment_lengths, dtype=np.uint64)
-    float_sample_rate = float(sample_rate_in_GHz)
-    time_array = np.arange(0, np.max(segment_lengths)/float_sample_rate, float_sample_rate) 
+    time_array = np.arange(np.max(segment_lengths), dtype=float) / float(sample_rate_in_GHz)
 
     return time_array, segment_lengths
--- a/qupulse/hardware/util.py
+++ b/qupulse/hardware/util.py
@@ -95,6 +95,7 @@ def get_sample_times(waveforms: Union[Collection[Waveform], Waveform],
         segment_lengths.append(rounded_segment_length)
 
     segment_lengths = np.asarray(segment_lengths, dtype=np.uint64)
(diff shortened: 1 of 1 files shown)


terrorfisch:
Thanks! Unfortunately we have systems which use 2.3GHz where 
```
a=np.arange(10000)/2.3
b=np.arange(0, 10000/2.3, 1/2.3)
np.abs(a-b).max()
Out[9]: 9.094947017729282e-13
```

I guess the expensive thing here are the allocations so one optimization is 
```
time_array = np.arange(np.max(segment_lengths), dtype=float)
time_array /= float(sample_rate_in_GHz)

In [32]: %timeit c = np.arange(10000, dtype=float); c /= 2.4
18.4 µs ± 949 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [33]: %timeit a = np.arange(10000)/2.4
25.6 µs ± 579 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [34]: %timeit b = np.arange(0, 10000/2.4, 1/2.4)
8.08 µs ± 69.8 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```


This function might be a good candidate for a decoration with `functools.lru_cache` or sample rate based caching. Or do you need the cold start performance?

```python
CACHE = {}

def _get_sample_times(n: int, f: float):
    arr = CACHE.get(f)    
    if arr is None or arr.size < n:
        arr = np.arange(n, dtype=np.float64)
        arr /= f
        CACHE[f] = arr
    elif arr.size > n:
        arr = arr[:n]
    return arr
```

peendebak:
@terrorfisch I updated the PR to use your variation. Using a cache would add some more performance, but we would have to make the array readonly to make it safe. 
Cold start performance would not be an issue for our use case.

PR #565, opened 2021-03-18
Add ConstantPulseTemplate

peendebak:
Add template for multi-channel pulses that are constant. These pulses are easier to create than a `TablePulseTemplate` and allow backends to optimize uploaded of these waveforms.

@terrorfisch

### DIFF
--- a/qupulse/_program/waveforms.py
+++ b/qupulse/_program/waveforms.py
@@ -8,7 +8,7 @@ import itertools
 from numbers import Real
 from abc import ABCMeta, abstractmethod
 from weakref import WeakValueDictionary, ref
-from typing import Union, Set, Sequence, NamedTuple, Tuple, Any, Iterable, FrozenSet, Optional, Mapping, AbstractSet
+from typing import cast, Union, Set, Sequence, NamedTuple, Tuple, Any, Iterable, FrozenSet, Mapping, AbstractSet
 import operator
 
 import numpy as np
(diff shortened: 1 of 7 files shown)


peendebak:
@terrorfisch Can you review this PR?

qupulse/pulses/constant_pulse_template.py
```diff
@@ -0,0 +1,134 @@
...
+from qupulse.parameter_scope import Scope
+
+
+__all__ = ["ConstantPulseTemplate"]
+
+
+class ConstantPulseTemplate(AtomicPulseTemplate):  # type: ignore
+
+    def __init__(self, duration: float, amplitude_dict: Dict[str, Any], identifier: Optional[str] = None,
+                 name: Optional[str] = None, measurements: Optional[List[MeasurementDeclaration]] = [], **kwargs: Any) -> None:
```
terrorfisch, line 30: Is there a reason to use name instead of identifier?

qupulse/_program/waveforms.py
```diff
@@ -243,6 +243,43 @@ def __repr__(self):
...
+
+    @property
+    def defined_channels(self) -> Set[ChannelID]:
+        """The channels this waveform should played on. Use
+            :func:`~qupulse.pulses.instructions.get_measurement_windows` to get a waveform for a subset of these."""
+
+        return {self._channel}
+
+    def compare_key(self) -> str:
+        return 'ConstantWaveform:' + cast(str, hash([self._duration, self._amplitude, self._channel]))
```
terrorfisch, line 267: Typically a tuple of hashables is returned here that is independent of the running instance since the result of `hash(obj)` should change with each interpreter restart.

qupulse/_program/waveforms.py
```diff
@@ -243,6 +243,43 @@ def __repr__(self):
...
 
 
+class ConstantWaveform(Waveform):  
+    def __init__(self, duration: float, amplitude: Any, channel: ChannelID):
+        """ Create a qupulse waveform from a qupulse template that has a QI waveform inside """
+        self._duration = duration
+        self._amplitude = amplitude
+        self._channel = channel
+
+        self._is_constant_waveform = True
```
terrorfisch, line 253: What is the purpose of this field?

qupulse/_program/waveforms.py
```diff
@@ -243,6 +243,43 @@ def __repr__(self):
         return f'{type(self).__name__}(channel={self._channel_id}, waveform_table={self._table})'
 
 
+class ConstantWaveform(Waveform):  
+    def __init__(self, duration: float, amplitude: Any, channel: ChannelID):
+        """ Create a qupulse waveform from a qupulse template that has a QI waveform inside """
```
terrorfisch, line 248: Is there a direct relation to QI?

terrorfisch (commented):
I think it is a good idea to specialcase constant waveforms and is it something which is on my bucket list aswell.

I think `ConstantWaveform` tests are missing.

qupulse/_program/waveforms.py
peendebak: It is used by our backend (for the HDAWG8) to get a qupulse template represents a constant pulse/waveform.

qupulse/pulses/constant_pulse_template.py
peendebak: I believe the identifier should be unique the moment the template is serialized. Since several of my pulses have the same name, I added the `name`

qupulse/_program/waveforms.py
terrorfisch: I am thinking about adding some inspection properties such as `is_constant` and `is_zero` in the mid-term. Until then this is a good solution, I think.

terrorfisch (approved):

qupulse/pulses/constant_pulse_template.py
terrorfisch: Ah, so its more a human readable tag? What do you use it for? Do you think it makes sense to extend this to other pulse templates?

terrorfisch:
Looks good. Thank you for the contribution! :)

One last thing. Could you add the ConstantPT to ReleaseNotes.txt?

eendebakpt:
> Looks good. Thank you for the contribution! :)
> 
> One last thing. Could you add the ConstantPT to ReleaseNotes.txt?

Done!

PR #583, opened 2021-05-12
Add AtomicSequencePulseTemplate

peendebak:
Add pulse template similar to the `SequencePulseTemplate`, but which acts as an atomic pulse template.

@terrorfisch

### DIFF
--- a/qupulse/pulses/atomic_sequence_pulse_template.py
+++ b/qupulse/pulses/atomic_sequence_pulse_template.py
@@ -1,0 +1,254 @@
+"""This module defines the AtomicSequencePulseTemplate
+
+The AtomicSequencePulseTemplate is similar to the SequencePulseTemplate, but acts as an atomic template.
+This allows definitions of pulses that are combined of pulses that cannot be played by a backend 
+(e.g. because of restrictions on the number of samples)
+
+Classes:
+    - AtomicSequencePulseTemplate: Defines a pulse
+"""
+
+import unittest
+import functools
+import logging
+import numbers
+import warnings
+from functools import wraps
+from typing import Any, Dict, List, Optional, Sequence, Set, Tuple, Union
+
+import numpy as np
+
+from qupulse._program.waveforms import SequenceWaveform
+from qupulse.expressions import Expression, ExpressionScalar, Number
+from qupulse.parameter_scope import DictScope, Scope
+from qupulse.pulses.constant_pulse_template import ConstantPulseTemplate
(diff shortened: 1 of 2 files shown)


terrorfisch:
Do you have a usecase where the `to_single_waveform` of `create_program` argument does not work? I included it so you can translate arbitrary pulse template combinations to waveforms but I guess it is a bit tedious to use.

peendebak:
> Do you have a usecase where the `to_single_waveform` of `create_program` argument does not work? I included it so you can translate arbitrary pulse template combinations to waveforms but I guess it is a bit tedious to use.

The use case for us is the following. We have a standard initialization-manipulation-readout exeperiment. The initialization and readout and in the order of microseconds and we can round durations to fit with the AWG. The manipulation stage is more time critical, so we work with ns pulses and at the end wrap the whole manipulation stage in a `AtomicSequencePT`. 

Converting the whole sequence with `to_single_waveform` would not work, since that sequence is too long. (we need the pulses in the initialization stage to be optimized by the AWG backend)

@terrorfisch

terrorfisch:
My approach to this problem would be to use a `SequencePT` with an identifier so it doesnt get merged with the rest. It would then look like this

```python
experiment = initialization @ SequencePT(manipulation_1, manipulation_2, ..., identifier='manipulation_whatever') @ readout
program = experiment_pt.create_program(to_single_waveform={'manipulation_whatever'})
```

Alternatively one can create all `SequencePT`s explicitly and pass the pulse template since only the `@` operator merges them.
```python
manipulation = SequencePT(manipulation_1, manipulation_2, ...)
experiment = SequencePT(initialization, manipulation, readout)
program = experiment_pt.create_program(to_single_waveform={manipulation})
```

peendebak:
@terrorfisch I was not aware of that functionality in the `create_program`. It does mean that the right arguments (e.g. `to_single_waveform={manipulation}`) have to be passed to the location(s) where `create_program` is used. That does happen in the plotting routines, although there we don't care whether the pulse is atomic or not. The other location is where the pulses are rendered by the AWG driver.
I will think about whether this approach works for us.

terrorfisch:
Yes. The idea is to keep the hardware requirement that some pulse parts should be concatenated into a single waveform reasonably seperate from the pulse itself to make pulses more portable and be more flexible when rearranging them.

peendebak:
Closing this PR. We will keep the `AtomicSequencePulseTemplate` in our local codebase

PR #585, opened 2021-05-16
Issues/584 uninitialized table sample

### LINKED ISSUE #584: TablePT can generate random values
A minimal example is:
```
import matplotlib.pyplot as plt
from qupulse.pulses import TablePT 
from qupulse.pulses.plotting import plot

import qupulse.utils.sympy
import qupulse._program.waveforms

qupulse._program.waveforms.PULSE_TO_WAVEFORM_ERROR=1e-6


d=598.3333333333334-480
c2=TablePT(entries={'P': [(0, 1.0, 'hold'), (d, 1.0, 'hold')]})

plt.figure(2)
plt.clf()
_ = plot(c2, sample_rate=3., axes=plt.gca() )
```
The output is:

image

To show this properly in `qupulse.pulses.plotting.render` I changed
```
    voltages = {ch: np.empty_like(times)
                for ch in channels}
```
into
```
    voltages = {ch: np.empty_like(times)*0+1000
                for ch in channels}
```
so that the undefined values from `empty_like` are set to 1000.

The `qupulse._program.waveforms.PULSE_TO_WAVEFORM_ERROR=1e-6` introduces a rounding error where a sample is selected at a time slightly larger then the the duration of the `TablePT`. I could not quickly create a minimal example for the default `qupulse` settings.

@terrorfisch

terrorfisch:
Fixes #584

### DIFF
--- a/qupulse/_program/waveforms.py
+++ b/qupulse/_program/waveforms.py
@@ -20,6 +20,7 @@ from qupulse.comparable import Comparable
 from qupulse.expressions import ExpressionScalar
 from qupulse.pulses.interpolation import InterpolationStrategy
 from qupulse._program.transformation import Transformation
+from qupulse.utils import pairwise
 
 
 __all__ = ["Waveform", "TableWaveform", "TableWaveformEntry", "FunctionWaveform", "SequenceWaveform",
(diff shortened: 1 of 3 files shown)


qupulse/_program/waveforms.py
```diff
@@ -226,7 +229,14 @@ def unsafe_sample(self,
         if output_array is None:
             output_array = np.empty_like(sample_times)
```
peendebak, line 230:
```suggestion
            output_array = np.empty_like(sample_times)
            output_array.fill(np.NaN)
```
This PR fixes the example, but there might be other corner cases. This would make the problems more explicit.
(there are other places in the codebase where this applied as well)

peendebak (approved):

qupulse/_program/waveforms.py
terrorfisch: Probably a good idea. I added an `ALLOCATION_FUNCTION` for now to have the option to configure this with minimal overhead.

PR #600, opened 2021-06-08
Add automatic constant waveform optimization

terrorfisch:
Adds the methods `is_constant`, `constant_value_dict` and `constant_value` to `Waveform`. `constant_value` is used by `get_sampled` to shortcut. They can also be used by a hardware backend optimize program translation.

Waveforms are now atomatically created as constant waveforms if created via one of the classmethods `from_*`.

Example:
```python
from qupulse._program.waveforms import FunctionWaveform
print(
    FunctionWaveform.from_expression(Expression(4), channel='A', duration=3)
)
```

prints `<qupulse._program.waveforms.ConstantWaveform object at 0x00000285380B3388>`

TODO:
 - [x] Add newsfragment(s)
 - [ ] 100% Test coverage of new lines


terrorfisch:
@peendebak

qupulse/_program/waveforms.py
```diff
@@ -262,6 +327,7 @@ def __repr__(self):
 
 class ConstantWaveform(Waveform):
 
+    # TODO: remove
     _is_constant_waveform = True
```
terrorfisch, line 331: @peendebak can this be removed or would you rather keep it around?

eendebakpt: @terrorfisch Can be removed once I updated our stack. It is fine to deprecate it right away

PR #617, opened 2021-11-16
Travis CI maintanance

terrorfisch:
- Use Ubuntu 20.04 LTS and drop hard gcc-8 dependency.
 - Call pytest directly instead of the deprecated setup.py test.
 - Add python 3.9 to test suite


terrorfisch:
We use github actions now

PR #626, opened 2021-12-29
improve efficiency of TablePulseTemplate integral

peendebak:
This PR improves the efficiency of calculation of integrals for the TablePulseTemplate. If there are no free variables in the TablePulseTemplate this calculation can be done even more efficiently, but this is not implemented in this PR.

- Lambdify integral expressions
- Use class variables for the expression and integral
- Reduce double evaluations in the TableEntry._sequence_integral

@terrorfisch


terrorfisch:
Thanks for the contribution! Looks reasonable.

I will take this as an incentive to finally repair the CI testing and afterwards take a closer look.

PR #630, opened 2022-01-26
[Performance] Optimize waveform_duration

peendebak:
The `waveform.duration` is an expensive property (due to the type used). This PR eliminates some double evaluations.

@terrorfisch

PR #631, opened 2022-01-27
Improve performance of Waveform class

terrorfisch:
- Make it a `__slots__` class for lower memory footprint.
 - Calculate `duration` upon construction.


terrorfisch:
@peendebak This should make #630 obsolete and improve performance even more. Do you have a test case at hand where you can try it?

peendebak:
@terrorfisch The PR works. Performance does not increase even more (I guess the bottlenecks are elsewhere).

terrorfisch:
Sadly at some point python itself is the limiting factor.

PR #637, opened 2022-02-09
Cast np.float64 to float in ExpressionScalar initialization

peendebak:
@terrorfisch


terrorfisch:
Thanks for the headsup. I am currently rewriting the expressions ( #640 ) and will implement it there. Isnt it enough to cast to float in `__float__`?

peendebak:
@terrorfisch Casting in `__float__` should be fine as well

PR #640, opened 2022-02-11
Rewrite Expression code and use symengine in some cases

terrorfisch:
First draft of `Expression` rewrite to gauge if the performance gains are worth it.

TODO:
 - [ ] Benchmark
 - [ ] Remove old expression code
 - [ ] More tests

### DIFF
--- a/qupulse/_program/waveforms.py
+++ b/qupulse/_program/waveforms.py
@@ -466,7 +466,7 @@ class FunctionWaveform(Waveform):
                       channel: ChannelID,
                       sample_times: np.ndarray,
                       output_array: Union[np.ndarray, None] = None) -> np.ndarray:
-        evaluated = self._expression.evaluate_numeric(t=sample_times)
+        evaluated = self._expression.evaluate_in_scope(FrozenDict(t=sample_times))
         if output_array is None:
             if self._expression.variables:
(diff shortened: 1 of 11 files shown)


terrorfisch:
This made things slower, going from

599 ms to 824 ms in my test case...

peendebak:
> ~This made things slower, going from~ Edit: Had it the wrong way around. It gets faster.
> 
> 599 ms vs. 824 ms in my test case.
> 
> ```python
> from qupulse.pulses import *
> ppt = PointPT([(0, 'i * step + start'), (192, 'i')], ('a',))
> fpt = ForLoopPT(ppt, 'i', 'n')
> program = fpt.create_program(parameters=frozendict(n=10000, step=0.01, start=-.4))
> 
> %timeit fpt.create_program(parameters=frozendict(n=10000, step=0.01, start=-.4))
> ```

@terrorfisch Performance gain on my system is similar you what you report here. Not sure whether the gain is from symengine or not (if available, sympy can use symengine as a backend already). Some profiling shows that pend in the sympy/symengine methods is not dominant

image

image

(generated with https://jiffyclub.github.io/snakeviz/)

terrorfisch:
Requries complete rework anyways

PR #652, opened 2022-05-12
Implement __eq__ for Serializables

terrorfisch:


terrorfisch:
@thangleiter should work now

PR #661, opened 2022-06-24
Improve performance of expressionscalar operations

eendebakpt:
For operators like `>` or `==` it is more efficient to let sympy do the conversion of non-sympy types. For types like `int` and `float` sympy can optimize.

@terrorfisch


terrorfisch:
Would probably be even faster to compare with `getattr(other, '_sympified_value', other)`

eendebakpt:
> Would probably be even faster to compare with `getattr(other, '_sympified_value', other)`

Done

PR #664, opened 2022-06-30
Add MethodPulseTemplate

eendebakpt:
The `MethodPulseTemplate` allows to define a method using any Python function that accepts a time parameter (as a numpy array) and returns the calculated values of the waveform.

@terrorfisch

### DIFF
--- a/qupulse/pulses/method_pulse_template.py
+++ b/qupulse/pulses/method_pulse_template.py
@@ -11,7 +11,7 @@ from qupulse.serialization import PulseRegistryType, Serializer
 
 
 import functools
-@functools.lru_cache
+@functools.lru_cache(maxsize=1024)
 def ExpressionScalarCache(value):
     return ExpressionScalar(value)
         


terrorfisch:
I am conflicted about the seriazability. Do you have specific use cases for this?

eendebakpt:
@terrorfisch No, I added it because i thought every template should be able to do this. We can also remove it

terrorfisch:
I totally see that this can be usefull to easily hack in a desired behaviour that cannot be done with FunctionPT.
The downside is that it makes it very easy to build a non-serializable and environment dependent pulse template. You can already do this now, for example by modifying the sympy namespace but then it's much clearer to the implementer that the pulse is not portable. Therefore I would keep it out for now until we have a usecase.

eendebakpt:
> I totally see that this can be usefull to easily hack in a desired behaviour that cannot be done with FunctionPT. The downside is that it makes it very easy to build a non-serializable and environment dependent pulse template. You can already do this now, for example by modifying the sympy namespace but then it's much clearer to the implementer that the pulse is not portable. Therefore I would keep it out for now until we have a usecase.

To be clear: I do not have a use case for the serialization, but I do have a use case for the `MethodPulseTemplate`. For me it is both faster and more convenient than the `FunctionPT`. But ok to leave it out for now. With the rust backend the performance considerations will probably change.

terrorfisch:
Ah, I misunderstood you than. Can you share what kind of functions you want to use this with?

eendebakpt:
> Ah, I misunderstood you than. Can you share what kind of functions you want to use this with?

This is the equivalent of what I am using with a `FunctionPT`. It is a basic IQ channel definition.

```
from qupulse.pulses import MappingPT, FunctionPT
from qupulse.pulses.multi_channel_pulse_template import AtomicMultiChannelPulseTemplate
from qupulse.pulses.plotting import plot as plot_pulse
from qupulse.pulses.plotting import render
from qtt.utilities.visualization import get_axis

amplitude=1.
f=116e6/1e9
phase=1.1
duration=100
ptI = FunctionPT('(exp(-(t-duration/2)^2/sigma^2) - exp(-(duration/2)^2/sigma^2))*cos(2*pi*freq*t+phi)', 100, channel='I')
ptQ = FunctionPT('(exp(-(t-duration/2)^2/sigma^2) - exp(-(duration/2)^2/sigma^2))*sin(2*pi*freq*t+phi)', 100, channel='Q')

pt2=MappingPT( AtomicMultiChannelPulseTemplate(ptI, ptQ) , parameter_mapping={'duration': duration, 'phi': phase, 'freq': f, 'sigma': duration/4 })

plot_pulse(pt2, axes=get_axis(10))


#%timeit render(pt2.create_program()) 
# 4.92 ms ± 419 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

#from sqt.measurements.pulse_generator import sample_pulse
#%timeit sample_pulse(pt2, 10) # sample a single data point
# 4.57 ms ± 314 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```
Rendering either the full template, or sampling a single data point takes about 5 ms. With the `MethodPulseTemplate` this is about 0.35 ms.

For the `FunctionPT` sime is spend in the `create_program`, but mostly in the sympy `lambdify` function.

PR #680, opened 2022-09-21
Remove override of _parse_evaluate_numeric_result

eendebakpt:
@terrorfisch With the new rust backend performance has improved, but off course there are new bottlenecks.

This PR removes a small bottleneck. I could not find a reason for the additional transformation in the current codebase, but perhaps you know why it was added.

### DIFF
--- a/qupulse/expressions.py
+++ b/qupulse/expressions.py
@@ -404,15 +404,15 @@ class ExpressionScalar(Expression):
     def is_nan(self) -> bool:
         return sympy.sympify('nan') == self._sympified_expression
 
-    def _parse_evaluate_numeric_result(self,
-                                       result: Union[Number, numpy.ndarray],
-                                       call_arguments: Any) -> Number:
-        """Overwrite super class method because we do not want to return a scalar numpy.ndarray"""
-        parsed = super()._parse_evaluate_numeric_result(result, call_arguments)
-        if isinstance(parsed, numpy.ndarray):
-            return parsed[()]
-        else:
-            return parsed
+    # def _parse_evaluate_numeric_result(self,
+    #                                    result: Union[Number, numpy.ndarray],
+    #                                    call_arguments: Any) -> Number:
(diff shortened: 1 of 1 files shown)


terrorfisch:
I remember adding it to convert scalar numpy arrays into actual scalar values which resolved a problem at some point.

eendebakpt:
> I remember adding it to convert scalar numpy arrays into actual scalar values which resolved a problem at some point.

I gathered something like that. But the code is not actually converting to scalar values (e.g. to a `float`):

image

terrorfisch:
It is for the following
```python
import numpy as np
x = np.array(1.)
print(f"{x!r} vs {x[()]!r}")
```
prints `'array(1.) vs 1.0'`

eendebakpt:
> It is for the following
> 
> ```python
> import numpy as np
> x = np.array(1.)
> print(f"{x!r} vs {x[()]!r}")
> ```
> 
> prints `'array(1.) vs 1.0'`

I know those things exist, but don't know why they are created in the first place. I modified the pr to add back the conversion, but now in the parent class.
That should be a bit more efficient, although I did not benchmark yet.

terrorfisch:
AFAIR this was in the context of the PointPulseTemplate. Non scalar values/expressions introduced a lot of problems...

eendebakpt:
@terrorfisch The current version of the PR still handles the conversion of numpy scalar arrays. But by doing it in the `Expression` class we reduce the number of `instance(..., np.ndarray)` checks from 2 to 1. By checking `result.shape` we avoid the conversion for normal (e.g. non-scalar arrays).

```
In [13]: import qupulse
    ...: import numpy as np
    ...: g=qupulse.expressions.ExpressionScalar(np.array(2.))
    ...: print(type(g.evaluate_numeric()))
<class 'float'>
```

terrorfisch:
I restructured everything a bit here: https://github.com/QuTech-Delft/qupulse/pull/5

PR #681, opened 2022-09-21
Faster check on monotomicity

eendebakpt:


terrorfisch:
I think there is a not missing here.

For completeness: https://github.com/qutech/qupulse/pull/503 Introduces a [significantly faster implementation](https://github.com/qutech/qupulse/pull/503/files#diff-88f1ebadb7189bd8391672a8d851b6a00ebab8104479193d83a498f20accfa37).

eendebakpt:
> I think there is a not missing here.
> 
> For completeness: #503 Introduces a [significantly faster implementation](https://github.com/qutech/qupulse/pull/503/files#diff-88f1ebadb7189bd8391672a8d851b6a00ebab8104479193d83a498f20accfa37).

Bug was fixed. The numpy implementation in #503 is similar to this PR. The numba version is even faster:
```
In [27]: %timeit _is_monotonic_numpy(sample_times)
5.95 µs ± 480 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [28]: %timeit _is_monotonic_numba(sample_times)
2.44 µs ± 264 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```
Are you  still considering numba if there is also a rust backend in the making? With pyo3 a small function to check monotonicity in rust would be just as fast (or faster) than the numba version.

PR #685, opened 2022-09-30
Feat/mfli driver

terrorfisch:


paulsurrey:
@terrorfisch only for testing things.

terrorfisch:
@paulsurrey should we focus on moving this forward?

terrorfisch:
What do you think about utilizing https://github.com/mwouts/jupytext/ to manage the examples?

terrorfisch:
I think we should create a separate repository for this in the spirit of #779

terrorfisch:
Will be moved to a seperate repository

PR #689, opened 2022-10-10
Fix defined channels in ConstantPT

eendebakpt:


terrorfisch:
I will merge this now to fix master but I think we should migrate to `AbstractSet` in the long run to allow for more optimized implementations of necessary.

PR #742, opened 2023-01-09
Add files via upload

daniel204:
Added the DecaDAC driver. Ignore all the commented blocks. It's mostly there in case I want to reverse some of the changes I've made. I'll delete these in a later stage

### DIFF
--- a/qupulse/hardware/awgs/decadac.py
+++ b/qupulse/hardware/awgs/decadac.py
@@ -1,0 +1,440 @@
+from abc import ABC
+from numbers import Real
+from typing import Tuple, Callable, Optional, Sequence, Union, Dict, Mapping, Set
+import serial
+
+from qupulse import ChannelID
+from qupulse._program._loop import Loop, to_waveform
+from qupulse.hardware.awgs.base import AWG
+
+'''upload general plan:
+    get timeline, with each instance being an index in an array
+    convert every voltage value to a decadac value
+    look for ramps '''
+def convert_voltage(v) -> int:
+    return int((v + 10) / 20 * 65535)
+
+#TODO delete this at one point
+'''class Single:
+
+    def __int__(self, voltage, duration):
+        self.d = voltage
+        self. duration = duration
+
+    def get_ascii(self):
+        return 'D' + str(self.d)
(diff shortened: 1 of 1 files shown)


qupulse/hardware/awgs/decadac.py
```diff
@@ -0,0 +1,440 @@
...
+        
+class Assign:
+
+    def __init__(self, voltage, channel: tuple):
+        self._voltage = voltage
+        self._channel = channel # in the form of (B,C) where B is the board and C is the channel. Possible values: B 0-4, C 0-3
+
+
+    def get_ascii(self):
+        command = 'B' + str(self._channel[0]) + ';C' + str(self._channel[1]) + ';'
```
terrorfisch, line 120: Use the python format string literals where possible. It makes reading the ASCII commands in the code easier and is generally faster.
  ```python
command = f"D{self._channel[0]};C{self._channel[1]};"
# vs
command  = 'B' + str(self._channel[0]) + ';C' + str(self._channel[1]) + ';'
# although the best would be
command = f"D{self.block};C{self.channel};"
```

File: qupulse/hardware/awgs/decadac.py [hunk 2]
```diff
@@ -0,0 +1,440 @@
...
+class DecaDACAWG(AWG):
+
+    def __init__(self, port: str, identifier='DecaDAC',):
+        super().__init__(identifier=identifier)
+
+        if port is not str:
+            raise TypeError(
+                'Declared port must be a string. Usually in the form of COM + n, with n being the port number')
+        self._ser = serial.Serial(port, 9600, timeout=0)
+        self._channel_voltage_location = {1: 1545,
```
terrorfisch, line 201: Put this information in a global constant with an all capital name. And it looks like the formula is

```python
VOLT_LOCATION_OFFSET = 1545
def get_channel_voltage_location(channel: int) -> int:
    return VOLT_LOCATION_OFFSET + channel * 6
```

terrorfisch (commented):
Overall code structure looks good :)

I made a few remarks but I guess You'll need to change some of it anyways for the code to work.

PR #765, opened 2023-05-23
ParallelConstantChannelPT(...).defined_channels typo

paulsurrey:
The set.union call within the property definition throws an error:

```
File c:\users\lablocal\documents\git\qupulse\qupulse\pulses\multi_channel_pulse_template.py:256, in ParallelConstantChannelPulseTemplate.defined_channels(self)
    254 @property
    255 def defined_channels(self) -> Set[ChannelID]:
--> 256     return set.union(self._template.defined_channels, self._overwritten_channels.keys())

TypeError: descriptor 'union' for 'set' objects doesn't apply to a 'dict_keys' object
```

This can be fixed by casting the dict_keys into sets.

**Note** that this branch is branched of the feat/mfli_driver branch. Which is not to be merged.


terrorfisch:
Please cherry pick the commit.

PR #772, opened 2023-06-23
Add a simple implementation of the ELFManager that is used by default

### LINKED ISSUE #751: HDAWG: Multiple issues during .elf file upload
There are multiple issues present with the upload of .elf files. In the default case that auto-upload is enabled, those issues only manifest in error messages, but the upload itself works. With auto-upload disabled no error is thrown but .elf files are also not uploaded. However, this is compensated by another bug which makes it difficult to actually deactivate auto-upload.

## Auto-Upload enabled
The following lines:
~~~python
HardwareSetup.remove_program("playground_program")
HardwareSetup.register_program("playground_program", pulse_program, run_callback=awg.channel_tuples[0].run_current_program)

HardwareSetup.run_program("playground_program")
~~~

Cause the following error:
~~~bash
...
File successfully uploaded
Uploading ELF file to device DEV8212, core 3
File successfully uploaded

Output exceeds the [size limit](command:workbench.action.openSettings?[). Open the full output data [in a text editor](command:workbench.action.openLargeOutput?c0865e13-c6e5-4064-8c7e-461877ce63d8)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
~AppData\Local\Temp\ipykernel_21684\1993656821.py in <module>
      9 my_awg.awg_module.set('compiler/upload', 1)
     10 
---> 11 HardwareSetup.run_program("playground_program")

c:\users\beer\repos\qupulse\qupulse\hardware\setup.py in run_program(self, name)
    221     def run_program(self, name) -> None:
    222         """Calls arm program and starts it using the run callback"""
--> 223         self.arm_program(name)
    224         self._registered_programs[name].run_callback()
    225 

c:\users\beer\repos\qupulse\qupulse\hardware\setup.py in arm_program(self, name)
    212         for awg in self.known_awgs:
    213             if awg in awgs_to_upload_to:
--> 214                 awg.arm(name)
    215             else:
    216                 # The other AWGs should ignore the trigger

c:\Users\Beer\venvs\qupulse\lib\site-packages\autologging.py in autologging_traced_instancemethod_delegator(self_, *args, **keywords)
   1039             return proxy(method, args, keywords)
   1040         else:
-> 1041             return method(*args, **keywords)
...
-> 1007             _, status = self._upload_job
   1008             if isinstance(status, int):
   1009                 assert status in (-1, 0, 1)

TypeError: cannot unpack non-iterable NoneType object
~~~

This is caused by `self._upload_job` in `zihdawg.py` still being at it's default value, being `None`. This can relatively easily be fixed by modifying `ELFManager._update_upload_job_status` while still keeping validation of successful upload in the follow way:

~~~python
...
 def _update_upload_job_status(self):
        elf_upload = self.awg_module.elf_upload
        elf_file = self.awg_module.elf_file
        old_status = None

        if self._upload_job is None:
            assert not elf_upload
        else:
            elf_file, old_status = self._upload_job
            assert self.awg_module.elf_file == elf_file
        
        if isinstance(old_status, float) or old_status is None:
            status_int, progress = self.awg_module.elf_status
...
~~~

While this is not changed the problem can be circumvented on the user side by first arming the program, which works but throws the same error, which can now be caught, and then running the program:

~~~python
try:
    HardwareSetup.arm_program("playground_program")
except TypeError:
    print("Catch Type error due to bug when using auto upload")

HardwareSetup.run_program("playground_program")
~~~

I am unsure while this error does not occur for other users. If my solution is satisfactory I can create a pull request to deploy the fix.

## Auto-Upload disabled
More problematic is what happens when auto-upload is disabled, which can be seen by this:

~~~python
HardwareSetup.remove_program("playground_program")
HardwareSetup.register_program("playground_program", pulse_program, run_callback=awg.channel_tuples[0].run_current_program)

for awg_i in HardwareSetup.known_awgs:
    my_awg = awg_i

my_awg.awg_module.set('compiler/upload', 0)

HardwareSetup.run_program("playground_program")
~~~

not causing an error on program run. This is a problem however because it turns out that the the program is not actually uploaded. This can easily be checked by either observing no output signal from the awg or by checking that `sequence on devXXXX` in LabOne stays empty. This is caused by the following:

Upon this:
~~~python
    def _upload(self, elf_file) -> Generator[str, str, None]:
        if self.awg_module.compiler_upload:
            pass
        else:
            self._start_elf_upload(elf_file)
...
~~~

Manual upload is supposed to be triggered if `compiler/upload` is 0 by `zihdawg._start_elf_upload(elf_file)`. In there it is checked wehter the program is already uploaded in order to not perform unnecessary uploads by the following:

~~~python
 def _start_elf_upload(self, elf_file):
        ...
        current_elf = self.awg_module.elf_file
        if current_elf != elf_file:
        ... (start upload)
~~~

The problem is that `current_elf != elf_file` is always `False` as `elf/file` does not return the *uploaded* file but rather the *compiled* one as `current_elf`. This can be seen in the docs:

Screenshot 2023-02-02 105835

As `elf_file` is set when instructing the ZIdriver to compile the file, these values are always equal (see top-left corner):

Screenshot 2023-02-02 110032

It does not appear possible to me to get the actually uploaded .elf file to check for successful upload. Note that this must be possible in some way as apparently LabOne can do this as the `sequence on devXXXX`-tab suggests.

Possible solutions:
- Disable manual-uploads and fail if aut-upload is disabled with an error
- Always upload whenever the file that is compiled changes or is set for the first time

## The reason while this did not occur earlier:

It appears to be impossible to disable Auto Upload from the LabOne interface in `20.07`, the Auto Upload button simply does nothing. Unsetting it resets to be set on page refresh (which is different to other settings) and changing it's state has no influence on `compiler/upload`, it is always `1` if not set from the API. Setting it to `0` from the API does also have no influence on LabOne while the result is apparently registered and applied in the device itself. Therefore Auto-Upload is always enabled for almost all users.

terrorfisch:
(Hopefully) fixes #751


terrorfisch:
@maxbeer99 @BrllsRX Can you test if this works on a live system?

terrorfisch:
I will merge this now since there have been no complaints and the current master version is definitively faulty

PR #787, opened 2023-08-23
Optimize unsafe_sample

eendebakpt:
In `unsafe_sample` the `sample_times` is guaranteed to be a numpy array. Calling `searchsorted` from the array is more efficient.

For a small numpy array:
```
In [66]: %timeit sample_times.searchsorted( (0,0), 'left')
1.53 µs ± 63.9 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [67]: %timeit np.searchsorted(sample_times, (0,0), 'left')
2.73 µs ± 424 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

@terrorfisch

### DIFF
--- a/qupulse/program/waveforms.py
+++ b/qupulse/program/waveforms.py
@@ -369,8 +369,8 @@ class TableWaveform(Waveform):
             entries = self._table
 
         for entry1, entry2 in pairwise(entries):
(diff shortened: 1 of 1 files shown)


terrorfisch:
Thanks. Seems like numpy's `array_function_dispatch` introduces quite some overhead.

PR #796, opened 2023-09-28
Tabor: Unique WF upload hack

Nomos11:


terrorfisch:
If it works it works. The downside is that the code gets harder to test the closer you get to the instrument. I remember that writing and debugging the changed code was quite unpleasant.

Nomos11:
Completed with #797

PR #797, opened 2023-09-28
Tabor segment deduplication

### LINKED ISSUE #795: Tabor: redundant waveform upload
When uplading a program which for some reason has multiple segments of equal waveforms in a `TaborProgram`, the `_find_place_for_segments_in_memory`-function does not check for unique values but uploads all segments from this batch of segments regardless. 
It seems to me thast this may be preventable by checking for unique hashes. Is this assumption correct?

terrorfisch:
Should fix #795 

 - [x] test with simulator

### DIFF
--- a/qupulse/_program/tabor.py
+++ b/qupulse/_program/tabor.py
@@ -430,7 +430,7 @@ class TaborProgram(ProgramEntry):
         else:
             self.setup_advanced_sequence_mode()
 
-        self._sampled_segments = self._calc_sampled_segments()
+        self._sampled_segments, self._waveform_to_segment = self._calc_sampled_segments()
 
     @property
     def markers(self) -> Tuple[Optional[ChannelID], Optional[ChannelID]]:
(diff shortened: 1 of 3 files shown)


terrorfisch:
@Nomos11 Can you test if this works?

Nomos11:
so far it seems to give me an IndexError at `to_upload_size = np.sum(new_segment_lengths[unknown] + 16)`, which implies that it still tries to find places for all redundant waveforms, I'll investigate that further

Nomos11:
It seems like the segment_lengths need to be adapted to throw out duplicates, which should be fixable relatively easily

terrorfisch:
Does it work now?

Nomos11:
From looking at the code I am a bit confused - does cutting off the array choose the correct values? the segment_lengths can contain subsequent duplicate values, which are not thrown out here, giving the incorrect values to my understanding

terrorfisch:
You are right. I did not think it through.

terrorfisch:
@Nomos11 Should work now

PR #805, opened 2023-10-13
Exemplary to_next_multiple

Nomos11:
could also be put somewhere else.

(Tests won't work here because floordiv not yet implemented in this branch)

### DIFF
--- a/qupulse/utils/__init__.py
+++ b/qupulse/utils/__init__.py
@@ -7,8 +7,6 @@ import numbers
 from collections import OrderedDict
 from frozendict import frozendict
 from qupulse.expressions import ExpressionScalar, ExpressionLike
-from qupulse.utils.types import TimeType
-from sympy import Max, sign
 
 import numpy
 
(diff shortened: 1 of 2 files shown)


qupulse/utils/__init__.py
```diff
@@ -122,3 +125,32 @@ def forced_hash(obj) -> int:
...
+    """
+    sample_rate = ExpressionScalar(sample_rate)
+    #is it more efficient to omit the Max call if not necessary?
+    if min_quanta is None:
+        #double negative for ceil division.
+        return lambda duration: -(-(duration*sample_rate)//quantum) * (quantum/sample_rate)
+    else:
+        #still return 0 if duration==0
+        return lambda duration: ExpressionScalar((quantum/sample_rate)\
+                                    *Max(-(-(ExpressionScalar(duration)*sample_rate)//quantum),min_quanta)\
```
terrorfisch, line 154: We should try to avoid relying on sympy internals.

Nomos11: Somehow a functionality that evaluates the min_quanta within the Expression has to be implemented. I did not directly see a different way than to use some sympy-functionality within the expression. Or are you just referring to the fact that this should be written out in e.g. a string to initialize the ExpressionScalar with?

terrorfisch: `f'max({min_quanta}, {ceiled_division(duration, quantum)}) * {quantum}'` could work although I am not sure how robust the interaction with TimeType is.

Nomos11: This seems to indeed be a bit problematic with TimeType (the tests give me weird behavior at the moment with a similar string), although I do not fully understand why.

Nomos11: 7c520a7 might resolve this; at least the test cases seem to work

PR #810, opened 2024-03-08
preliminary update for new firmware of HDAWG

Nomos11:


terrorfisch:
We shall move this into qupulse-hdawg

terrorfisch:
Moved to https://github.com/qutech/qupulse-hdawg

PR #820, opened 2024-04-15
compare hdawg_linspace hacky changes to main

Nomos11:
not commented/documented


tests/program/linspace_tests2.py
```diff
@@ -0,0 +1,289 @@
...
+    def test_increment_commands(self):
+        commands = to_increment_commands([self.program])
+        self.assertEqual(self.commands, commands)
+
+    def test_repeated_increment_commands(self):
+        commands = to_increment_commands([self.repeated_program])
+        self.assertEqual(self.repeated_commands, commands)
+
+
+class SingletLoadProcessing():
```
terrorfisch, line 171: This test is no executed. Is this intentional?

tests/program/linspace_tests2.py
terrorfisch, line 1: The filename has to end in `_tests.py` to be discovered (I think)

qupulse/hardware/awgs/base.py
```diff
@@ -162,17 +166,26 @@ def __str__(self) -> str:
...
+    FSP = -1
+    Loop = 0
+    Linspace = 1
+
+
 class ProgramEntry:
     """This is a helper class for implementing awgs drivers. A driver can subclass it to help organizing sampled
     waveforms"""
-    def __init__(self, loop: Loop,
+    def __init__(self, program: AllowedProgramTypes,
```
terrorfisch, line 180: I think it might make sense to create a new class

qupulse/hardware/awgs/base.py
```diff
@@ -162,17 +166,26 @@ def __str__(self) -> str:
                " Use force to overwrite.".format(self.name)
 
 
+AllowedProgramTypes = Union[Loop,Sequence[LinSpaceNode],]
+
+class _ProgramType(Enum):
+    FSP = -1
```
terrorfisch, line 172: Waht is FSP?

Nomos11: was a hacky way to introduce a "FixedStructureProgram" into the HDAWG (... a program with a fixed seqc-structure), the remaining necessary code would only be stored away in `qupulse_hdawg`

qupulse/hardware/awgs/base.py
Nomos11: a new class for what specifically?

qupulse/program/linspace.py
```diff
@@ -143,6 +144,16 @@ def hold_voltage(self, duration: HardwareTime, voltages: Mapping[ChannelID, Hard
                 bases.append(value)
                 factors.append(None)
                 continue
+            #there mightbe a bug in some waveform.constant_value, where an array of float instead of float is returned
+            #(goes against the typehints)
```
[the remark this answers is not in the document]
Nomos11: okay, I'll remove the workaround

tests/program/linspace_tests2.py
Nomos11: probably not...

tests/program/linspace_tests2.py
Nomos11: the file unintentionally snuck in here anyway I believe, it's just a copy of the original linspace_tests for some local testing, will be deleted

qupulse/program/linspace.py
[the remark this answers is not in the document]
terrorfisch: This might be a bold move

qupulse/hardware/awgs/base.py
Nomos11: I do not know if this is explicitly necessary right now - at least to ensure compatibility with other awg-drivers, one only has to rename a few 'loop's to 'program' in function calls, since ProgramEntry is only really called internally. Then one can see from there if contructing something new is necessary...

qupulse/program/linspace.py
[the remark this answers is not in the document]
Nomos11: we'll notice when the quantum internet breaks because of this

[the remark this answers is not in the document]
Nomos11: probably doing something wrong, but it somehow it appears as if in the newest master-commit it still does not work

qupulse/program/transformation.py
```diff
@@ -328,7 +328,8 @@ def __call__(self, time: Union[np.ndarray, float],
 
     def _instantiated_values(self, time):
         scope = {'t': time}
-        return {channel: value.evaluate_in_scope(scope) if hasattr(value, 'evaluate_in_scope') else np.full_like(time, fill_value=value, dtype=float)
+        array_or_float = lambda x: np.full_like(time, fill_value=x, dtype=float) if hasattr(time, '__len__') and len(time)>1 else x
```
terrorfisch, line 331: We want a single element array if time is a single element array.

PR #826, opened 2024-05-29
Make loop / program relation clearer

terrorfisch:


terrorfisch:
@Nomos11 Was there a reason for the separate `__loop` and `_program` attributes?

Nomos11:
> @Nomos11 Was there a reason for the separate `__loop` and `_program` attributes?
No particular reason except trying to change as little old functionality as possible.

PR #828, opened 2024-05-31
change simpleexpression for equality handling

Nomos11:


terrorfisch:
Does SimpleExpression implement the expression protocol? If not we should definitely mention that in the docs.

Nomos11:
Not an expert there but does not seem so; moving it around was also just to avoid some circular imports at some point, it can also be moved further

terrorfisch:
This PR has a lot of unrelated changes

Nomos11:
yes; it's evolved to collecting random changes to merge into some other side branch to work on, not into master; thought that it would not require spamming more PRs for that

qupulse/program/transformation.py
```diff
@@ -88,7 +88,16 @@ def get_constant_output_channels(self, input_channels: AbstractSet[ChannelID]) -
 
 class ChainedTransformation(Transformation):
     def __init__(self, *transformations: Transformation):
-        self._transformations = transformations
+        #avoid nesting also here in init to ensure always flat hierachy?
```
terrorfisch, line 91: This is the job of the function `chain_transformations`

Nomos11: yes; basically copied over that code since the function instantiates a new ChainedTransformation.
I was'nt sure if ChainedTransformation is called directly anywhere instead of being instantiated with .chain, so I did the check again in the init function. If the class is not supposed to be initialized directly, it could also be removed again.

qupulse/program/linspace.py
```diff
@@ -24,45 +146,108 @@ class DepKey:
...
     waveform: Waveform
-    channels: Tuple[ChannelID, ...]
+    
+    def dependencies(self):
+        return {}
+    
 
+@dataclass
+class LinSpaceArbitraryWaveformIndexed(LinSpaceNodeChannelSpecific):
+    """This is just a wrapper to pipe arbitrary waveforms through the system."""
```
terrorfisch, line 221: I guess this is no longer true

qupulse/program/linspace.py
```diff
@@ -106,43 +295,61 @@ class LinSpaceBuilder(ProgramBuilder):
...
+        self._to_stepping_repeat = to_stepping_repeat
 
     def _root(self):
         return self._stack[0]
 
     def _get_rng(self, idx_name: str) -> range:
         return self._get_ranges()[idx_name]
 
-    def inner_scope(self, scope: Scope) -> Scope:
+    def inner_scope(self, scope: Scope, pt_obj: Optional['ForLoopPT']=None) -> Scope:
```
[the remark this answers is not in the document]
Nomos11: this is all hacked in right now to see if anything works with what i intended to do.
reasoning here to keep the glue code out of the pulse_template._internal_create_program

qupulse/program/linspace.py
Nomos11: yes, docstrings are of course not adjusted atm

qupulse/program/waveforms.py
```diff
@@ -215,7 +220,17 @@ def reversed(self) -> 'Waveform':
         """Returns a reversed version of this waveform."""
         # We don't check for constness here because const waveforms are supposed to override this method
         return ReversedWaveform(self)
-
+    
+    @abstractmethod
+    def _compare_subset_key(self, channel_subset: Set[ChannelID]) -> Hashable:
```
terrorfisch, line 225: What is the reason for this method? The current implementation is worse than `hash(wf.get_subset_for_channels(channel_subset))`. If we need better subset analysis, we can improve waveform normalization.

Nomos11: this was done before the compare_key thing was reworked, i believe.
but your code would take into account channel names, right? this is what i tried to avoid there to just compare potential sample output.

Nomos11: this change however is also not relevant for the linspace thing and propagated to the branch from somewhere

Issue #824, opened 2024-05-13
Sample a non constant waveform at reduced sample rate


terrorfisch:

Nomos11:
yes, that sounds reasonable to me.
However, would the user then explicitly have to take care of the changed requirements for pulse granularity? That might run into some not so user friendly handling issues again if the time is not by chance compatible with all relevant powers of 2 (but maybe the pad_to_next_multiple or how it was called can help in some form)

terrorfisch:
Yes this is also somewhing that bothers me.

The best I can come up with is

````python
shuttle = FunctionPT('sin(t / t_dur)', duration='t_dur')
s1 = shuttle.with_mapping({'t_dur': '192 * 2**i'})
s2 = shuttle.with_mapping({'t_dur': '256 * 2**i'})
s3 = shuttle.with_mapping({'t_dur': '320 * 2**i'})

sweep = (s1 @ s2 @ s3).with_iteration(i, 10, 1000)
```

Nomos11:
In the end I think this is sufficient though. qupulse should probably enforce strict regulations and throw errors otherwise, and another layer on top [c/sh]ould allow potential inacurracies be handled.
maybe some convenience functions can be drafted to assist this.

terrorfisch:
In principle we can add timing requirements to `PulseMetaData` as well in the future to allow imprecise timing.

Nomos11:
additionally, something with rolling out loops that would not natively be supported by the linspacebuilder/the hardware could potentially be included.

terrorfisch:
`minimal_sample_rate` mixes physical information and setup information (depends on hardware filtering and interpolation)

PR #830, opened 2024-06-01
time extension via SequencePT

Nomos11:
somewhat against the design principles discussed in #824, but maybe somewhat convenient for `pad_to` related utility?

branch [https://github.com/qutech/qupulse/tree/feat/time_extension_pulse_template](/qutech/qupulse/tree/time_extension_pulse_template) is more elaborate, however, in which situation is it really needed to deviate from a sequencePT-style?


terrorfisch:
I am not convinced that encoding this in the type has advantages. A class method on sequencept could do the same.

Nomos11:
The main point of increased convenience would probably be the automated insertion into the program creation by overwriting the `_create_program` method with the added `self `into `to_single_waveform`

tests/pulses/time_extension_pulse_tests.py
```diff
@@ -59,4 +61,23 @@ def test_quantities(self):
         duration_extended = self.extend_both.duration.evaluate_in_scope(self.parameters)
         duration_summed = self.main_pt.duration.evaluate_in_scope(self.parameters)\
             + self.parameters['t_prior'] + self.parameters['t_posterior']
-        self.assertEqual(duration_extended, duration_summed)
\ No newline at end of file
+        self.assertEqual(duration_extended, duration_summed)
+        
+    def test_pad_to_usecase(self):
```
Nomos11, line 66: What I mean by the previous comment is basically this, which gives a - in my opinion - quite convenient way of quickly extend waveforms with shorter and longer hold times guaranteed to match the hardware requirements

qupulse/program/linspace.py
```diff
@@ -434,23 +434,20 @@ def to_waveform(program: Sequence[LinSpaceNode], channels: Tuple[ChannelID]) ->
...
+    sequence = []
+    for node in program:
+        if type(node)==LinSpaceArbitraryWaveform:
+            sequence += [node.waveform]
+        elif type(node)==LinSpaceHold:
+            assert node.duration_factors is None, 'NotImplemented'
+            assert all(factor is None for factor in node.factors), 'NotImplemented'
+            #the channels should be sorted accordingly so we can do this.
+            sequence += [MultiChannelWaveform(
+                [ConstantWaveform(node.duration_base,base,ch) for base,ch in zip(node.bases,channels)])]
```
terrorfisch, line 446: This should be `ConstantWaveform.from_mapping(node.duration, dict(zip(channels, node.bases)))`.

qupulse/program/linspace.py
```diff
@@ -434,23 +434,20 @@ def to_waveform(program: Sequence[LinSpaceNode], channels: Tuple[ChannelID]) ->
...
-        
-        sequenced_waveform = SequenceWaveform.from_sequence(
-            sequence
-            )
-        return sequenced_waveform
\ No newline at end of file
+    sequence = []
+    for node in program:
+        if type(node)==LinSpaceArbitraryWaveform:
+            sequence += [node.waveform]
```
terrorfisch, line 440: IMO bad style and `sequence.append(node.waveform)` should be preferred. The performance does not matter here but in general `append` is about twice as fast.

qupulse/program/linspace.py
Nomos11: makes sense, did not see that method

qupulse/program/linspace.py
Nomos11: indeed, internet says + is O(n^2) (was not aware of that), timing it shows quite a difference for long lists.

tests/pulses/time_extension_pulse_tests.py
terrorfisch: I think a utility function like this would be sufficient

```python
def extend(self, *, total_duration = None, prepend = None, append = None) -> SequencePT:
    none_count = total_duration is None + prepend is None + append is None
    if none_count != 1:
        raise ValueError("Exactly two arguments must be specified.")
    # Return a SequencePT
```

Nomos11: for that case of not concatenating the waveforms, yes; what about the single wf case? (Which i initially added this for)

qupulse/program/linspace.py
terrorfisch: I don't think it is O(n^2). `sequence = sequence + [node.waveform]` is `O(n^2)`. When using `+=` you are just paying the constant overhead of creating a temporary single element list, iterating it and deleting it.

tests/pulses/time_extension_pulse_tests.py
terrorfisch: Ah, but why the SequencePT? Why represent the start and end as PulseTemplates? I would either have a PT that represents the time extention *without* introducing ConstantPTs in the front and in the back OR use SequencePT and add an atomic flag to it which encodes potential atomicity in the PT itself.

Nomos11: seemed like the least programmatic effort to me, but that can of course be changed.

qupulse/program/linspace.py
```diff
@@ -194,9 +194,26 @@ def with_repetition(self, repetition_count: RepetitionCount,
...
+    @contextlib.contextmanager
     def new_subprogram(self, global_transformation: 'Transformation' = None) -> ContextManager['ProgramBuilder']:
-        raise NotImplementedError('Not implemented yet (postponed)')
+        
+        inner_builder = LinSpaceBuilder(self._idx_to_name)
+        yield inner_builder
+        inner_program = inner_builder.to_program()
+
+        if inner_program is not None:
+            #measurements not yet included
```
terrorfisch, line 206: The linspace builder ignores measurements. The idea is to have a dedicated `MeasurementBuilder` to extract them.

Nomos11: ok

PR #832, opened 2024-06-03
Move pytest config to pyproject and use default testpath

terrorfisch:


terrorfisch:
I cannot reproduce the failures locally right now

PR #836, opened 2024-06-04
Deprecate Comparable

terrorfisch:
Deprecate comparable because it is "unpythonic" and makes comparisons hard to follow.

Three subclasses have been changed:

 - `AWG` now directly implements an `id` based hash and comparison.
 - `Waveform` is now compared and hashed based on the `__slots__` and `_duration` of the base class by default. `TransformingWaveform` overrides that because it caches results for more efficient multi channel sampling.
 - `Transformation` implementations now provide there own `__eq__` and `__hash__`.

### DIFF
--- a/qupulse/hardware/awgs/base.py
+++ b/qupulse/hardware/awgs/base.py
@@ -12,7 +12,7 @@ from numbers import Real
 from typing import Set, Tuple, Callable, Optional, Mapping, Sequence, List, Union, NamedTuple
 from collections import OrderedDict
 from enum import Enum
-# from itertools import chain
+import warnings
 
 from qupulse.hardware.util import get_sample_times, not_none_indices
 from qupulse.utils.types import ChannelID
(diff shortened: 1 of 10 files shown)


terrorfisch:
@Nomos11 Do you thiunk this is reasonable?

qupulse/hardware/awgs/base.py
```diff
@@ -142,6 +141,13 @@ def sample_rate(self) -> float:
     def compare_key(self) -> int:
         """Comparison and hashing is based on the id of the AWG so different devices with the same properties
         are ot equal"""
+        warnings.warn("AWG.compare_key is deprecated since 0.11 and will be removed in 0.12",
+                      DeprecationWarning, stacklevel=2)
```
terrorfisch, line 145: The return value us missing here

qupulse/program/transformation.py
```diff
@@ -241,8 +277,16 @@ def get_input_channels(self, output_channels: AbstractSet[ChannelID]) -> Abstrac
     def get_output_channels(self, input_channels: AbstractSet[ChannelID]) -> AbstractSet[ChannelID]:
         return input_channels
 
+    def __eq__(self, other):
+        return isinstance(other, OffsetTransformation) and self._offsets == other._offsets
```
terrorfisch, line 281: Should be the same over all Transformations with appropriate reasoning.

```
self._attribute == getattr(other, '_attribute ', None)
isinstance(other, SomeTransformation) and self._attribute == other._attribute 
```

terrorfisch:
@Nomos11 please merge if you don't have objections

Nomos11:
I'm not authorized to merge into master.

Nomos11 (approved):

PR #843, opened 2024-06-10
Switch to hatch and hatchling

terrorfisch:
The original reasoning for this branch was to benefit from hatch's version management. However, forwarding the current version into the build script requires further hacking. Nevertheless, hatch makes other development tasks (drafting releases) easier.

 - [x] test hatch in readthedocs 

After this is merged the next steps are:
 - Proper VCS versioning for development installs
 - Run tests via hatch
 - Cleanup github actions


terrorfisch:
@Nomos11 Is this understandable upon reading the section in the README?

Nomos11:
I think people interested in the build process would be able to familiarize themselves with hatch and consult its documentation if unclear, if this is what you are asking

PR #855, opened 2024-06-20
Remove qctoolkit alias and MATLAB code

### LINKED ISSUE #841: Remove MATLAB code and qctoolkit redirection
There are no setups running anymore which use these.

Pro:
 - Simplification of build and isntallation process
 - Confusion of new users

Con:
 - This will make deserialization of old (REALLY OLD) format pulses impossible as far as I remember

shumpohl:
Closes #841


shumpohl:
@Nomos11 I think this can be merged

PR #862, opened 2024-06-21
Linspace bug follow up with no bugs

terrorfisch:


tests/program/linspace_tests.py
```diff
@@ -63,6 +63,81 @@ def test_output(self):
...
+             #Wait(duration=TimeType(1000000, 1)),
+             #LoopJmp(idx=0),
+             
+             LoopLabel(idx=1, count=199),
+             Set(channel=0, value=-0.4, key=DepKey(factors=())),
+             Wait(duration=TimeType(100000, 1)),
+             Increment(channel=0, value=0.01, dependency_key=DepKey(factors=(10000000,))),
+             Wait(duration=TimeType(1000000, 1)),
+             LoopLabel(idx=2, count=9),
+             #also here, an increment 0 may be helpful (at least to be able to force).
```
terrorfisch, line 109: The increment is not emitted due to an optimization. Why would it be helpful?

terrorfisch: I would go further and replace loops that are constant by waits

tests/program/linspace_tests.py
```diff
@@ -63,6 +63,81 @@ def test_output(self):
...
+             #here is what currently happens:
+             Set(channel=0, value=-1.0, key=DepKey(factors=(10000000,))),
+             Wait(duration=TimeType(1000000, 1)),
+             LoopLabel(idx=0, count=9),
+             Wait(duration=TimeType(1000000, 1)),
+             LoopJmp(idx=0),
+             #however, i think this is what should happen (maybe also with an additional "Set" before,
+             #which might cause complications if omitted in other contexts like AWG amplitude:
+             #LoopLabel(idx=0, count=10),
+             #Set(channel=0, value=-1.0, key=DepKey(factors=(10000000,))),
```
terrorfisch, line 99: This set statement is not required. because it is the same in every iteration and can be lifted out of the loop. Why should it happen? This is exactly one intended case the post != pre condition is supposed to detect and optimize.

tests/program/linspace_tests.py
Nomos11: for wait it's not helpful. Just for other things, like WF scale, the program would need to know about the register it has to choose, which may work smoother if the corresponding Set command with the relevant DepKey is directly in front of it. Or maybe not, I dont' know.

terrorfisch: I think a scaled play should have an dependency key.

tests/program/linspace_tests.py
terrorfisch: the thing is that it is required to do this twice because the initial state changes. I brooded over this quite a while I remember :D But it might still be that I am wrong...

Nomos11: I don‘t think you‘re wrong; just in the very pathological case of repeating a sequence of 1000 pulses, where one element is just „Set“ with a Set command, one may want to prefer to have this codeblock only once and does not want that perceived optimization of omitting the Set command but having to write the repeated sequence twice.
But I don‘t think this is very relevant.

shumpohl:
Lessens about being more careful when merging...

shumpohl:
@Nomos11 I think this can be merged

PR #868, opened 2024-07-01
Cleanup example notebooks and make them compile again

terrorfisch:
- Makes all example notebooks use the default matplotlib backend
 - Add hatch command to clean all metadata and reset execution state
 - Cleans all metadata and resets
 - Skip hardware dependent notebook execution for doc building


terrorfisch:
There are still undetereministic outputs in the notebooks which is probably a thing we have to accept.

terrorfisch:
Merging since only the upload of the test results faiöed

PR #881, opened 2025-01-15
complicated test for linspacebuilder

Nomos11:
Test is not finished yet as command structure will be more complicated possibly.

This does not even translate to commands for me however; error in
`assert len(self.iterations) == len(previous.iterations)`, `linspace.py:330`

I believe to reproduce one needs multiple channels, and an iteration over a sequence of two pulses, one of which has a dependent value of the outer iteration, the other is another iteration over some inner loop (possibly with the outer dependency on another channel or so).

### DIFF
--- a/qupulse/program/linspace.py
+++ b/qupulse/program/linspace.py
@@ -327,10 +327,23 @@ class DepState:
     iterations: Tuple[int, ...]
 
     def required_increment_from(self, previous: 'DepState', factors: Sequence[float]) -> float:
-        assert len(self.iterations) == len(factors)
-        assert len(self.iterations) == len(previous.iterations) or all(factor == 0
-                                                                       for factor in factors[len(previous.iterations):])
+        """Calculate the required increment from the previous state to the current given the factors that determine
+        the voltage dependency of each index.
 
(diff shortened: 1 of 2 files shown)


terrorfisch:
Thanks! I will look into it,

terrorfisch:
The problem is that the DepKey drops trailing zeros but some other logic does not account for that. However fixing the assert leads to correct outputs.

qupulse/program/linspace.py
```diff
@@ -327,8 +327,10 @@ class DepState:
     iterations: Tuple[int, ...]
 
     def required_increment_from(self, previous: 'DepState', factors: Sequence[float]) -> float:
-        assert len(self.iterations) == len(previous.iterations)
         assert len(self.iterations) == len(factors)
+        assert len(self.iterations) == len(previous.iterations) or all(factor == 0
+                                                                       for factor in factors[len(previous.iterations):])
+
```
terrorfisch, line 333: This fix is not final. I would rather have something more consistent

terrorfisch:
The only error that I found was the wrong assert. I would suggest you merge this and hunt down the next bug ;)

Issue #882, opened 2025-01-15
extend pad_to to pad_to_all_atomic_subtemplates


Nomos11:
to increase convenience even more, the pad_to could be called from the top-level-pt to extend all atomic subtemplates to adhere to awg-specifications.
This is e.g. outlined in branch `feat/linspace_measurements` and should be simple to implement
 https://github.com/qutech/qupulse/blob/6facb1f52229de7a6773e89a524a243a055b94ce/qupulse/pulses/pulse_template.py#L442

Potentially for 1.0?
In that regard however, the PT time-extended in this manner should by default be translated into a single waveform, otherwise the convenience gets lost again.

Nomos11:
also the padding should allow to create the pulse as a single waveform,otherwise the convenience is lost
maybe by inheriting from atomicPT?
https://github.com/qutech/qupulse/blob/6facb1f52229de7a6773e89a524a243a055b94ce/qupulse/pulses/time_extension_pulse_template.py#L22

shumpohl:
Everything AWG specific should go into `create_program` if possible. Can you give a usecase?

> In that regard however, the PT time-extended in this manner should by default be translated into a single waveform, otherwise the convenience gets lost again.

Do you mean the root PT? Where is then the difference to `pad_to(..., to_single_waveform=True)`?

Nomos11:
But is ```to_single_waveform=True``` available on the master branch currently? it seems that a sequencePT is created when kwargs are provided which does not take this kwarg. Probably I don't understand it right now.

The usecase would be that I could imagine the user assuming this "padding" functionality to fix the awg quanta issue, but right now (maybe I'm missing the decisive step) a SequencePT is created which by default will be translated to two waveforms unless specified explicitly in create_program's ```to_single_waveform```, which seems clumsy (?) for many pulses

shumpohl:
> But is `to_single_waveform=True` available on the master branch currently?

Touché. One neat hack would be to add `to_single_waveform` as a `PulseTemplate.__init__` key word argument. I will make an issue for that.

> The usecase would be that I could imagine the user assuming this "padding" functionality to fix the awg quanta issue by extending the "same PT" object that the function is called on, but right now (maybe I'm missing the decisive step) a SequencePT is created which by default will be translated to two waveforms unless specified explicitly in create_program's `to_single_waveform`, which seems clumsy (?) for many pulses

[...]

PR #889, opened 2025-02-14
Add to_single_waveform property to PulseTemplate

### LINKED ISSUE #578: Atomicity overhaul
Currently the atomicity is half-baked into the type system. A pulse template is atomic if it is always translated into a single waveform. It currently is possible to translate all templates to waveforms by passing them in the `to_single_waveform` argument of `create_program`. However they do not implement the interface of `AtomicPulseTemplate` i.e. have a `build_waveform` method.

Some "transforming" pulse templates are not subclasses of `AtomicPulseTemplate` but are expected to behave as such if their wrapped pulse template is atomic:

 - `MappingPulseTemplate`
 - `ArithmeticPulseTemplate`

This might lead to problems if they are used in `AtomicPulseTemplate`s that expect their subtemplates to be atomic like `AtomicMultiChannelPulseTemplate` and `ArithmeticAtomicPulseTemplate`. One example for this is here: https://github.com/qutech/qupulse/pull/577#issuecomment-826136949

This requires some design thought. ~~Maybe something like a `TransformingPulseTemplate` or `TransparentPulseTemplate`?~~

shumpohl:
This PR adds a `to_single_waveform` argument to `PulseTemplate` which makes it translate itself into a single waveform on program creation.

Furthermore these pulse templates for ward the kwarg:
 - `SequencePT`
 - `RepetitionPT`
 - `ForLoopPT`.
 - `MappingPT`

Closes #578 
Enables #882 

 - [x] Add kwarg to other pulse templates
 - [x] write test
 - [x] newspiece

Documentation is moved to #894

### DIFF
--- a/changes.d/578.feature
+++ b/changes.d/578.feature
@@ -1,0 +1,2 @@
+Add a ``to_single_waveform`` keyword argument to ``SequencePT``, ``RepetitionPT``, ``ForLoopPT`` and ``MappingPT``.
+When ``to_single_waveform='always'`` is passed the corresponding pulse template is translated into a single waveform on program creation.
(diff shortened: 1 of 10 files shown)


shumpohl:
@Nomos11 To you think this is reasonable?

shumpohl:
@Nomos11 plz merge

PR #902, opened 2025-07-01
Add more sophisticated padding support and helper functions for that

shumpohl:
A qupulsy replacement for https://github.com/qutech/qupulse/blob/feat/linspace_measurements_reversed_registers_divisor/qupulse/pulses/pulse_template.py#L637

Implements #882 

TODO:
 - [x] rest of documentation
 - [x] test
 - [x] newspiece
 - [x] handle identifier and registry

### DIFF
--- a/changes.d/882.feature
+++ b/changes.d/882.feature
@@ -1,0 +1,1 @@
+Add functions ``PulseTemplate.pad_selected_subtemplates_to`` for padding inner templates and ``PulseTemplate.with_mapped_subtemplates`` as implementation helper.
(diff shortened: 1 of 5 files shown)


Nomos11:
seems more general and nicer.


not directly related to this code but for the intend behind it:
Is it already ensured on the main branch that padding atomic templates will result in another atomic template?
I believe that was also one change on the feature branch that padding would optionally create an alibi-template that mimics atomicity. alternatively somehow with the pulsemetadata form the other PR?

PR #905, opened 2025-07-01
Replace dedicated PulseTemplate.to_single_waveform property with a metadata class

shumpohl:
The metadataclass is intended to track non-pulsy information like translation hints or tags than can be consumed further down the line.

I spend to much time writing a half-baked field declaration in an attempt to unify flexibility, IDE hints and useful error messages.


Nomos11:
> @Nomos11 If I see it correctly I can just remove the to_single_waveform kwarg, nobody uses it.

Do you mean other code or users? The latter have in my experience never really used it anyway; for the former possible interplay with #902 might be relevant? (padding to create single waveform)

PR #918, opened 2025-07-28
shift things, prepare move of modified LSB to qupulse_hdawg

Nomos11:


qupulse/hardware/awgs/base.py
```diff
@@ -226,39 +222,24 @@ def __init__(self, program: AllowedProgramTypes,
...
-            else:
-                raise NotImplementedError()
-                    
+            #!!! this formulation is also unfortunate as the channel transformations are
+            # not applied to the waveforms but only to 'Set'/'Increment'-like commands in LSB
+            waveforms_dict = program.get_waveforms_dict(channels,self._channel_transformations())
+            waveforms = waveforms_dict.keys()
         if waveforms:
-            self._waveforms = OrderedDict(zip(waveforms, self._sample_waveforms(waveforms)))
+            self._waveforms = OrderedDict(zip(waveforms,self._sample_waveforms(waveforms)))
```
shumpohl, line 233: We should stick to whitespace after comma as it is used by most if not all style guides.

qupulse/hardware/setup.py
```diff
@@ -98,6 +98,7 @@ def register_program(self, name: str,
                          program: Loop,
                          run_callback=lambda: None,
                          update: bool = False,
+                         channels: Sequence[ChannelID] = None,
```
shumpohl, line 101: This is backwards incompatible because `measurements` can be passed as a positional argument. (But I think nobody does that so it should be fine.) Prefer appending new arguments, ideally making them keyword only from the start.

Nomos11: right; then let's just swap

shumpohl (approved):

PR #919, opened 2025-07-29
trafos handling DLV

Nomos11:


qupulse/program/transformation.py
```diff
@@ -63,7 +64,11 @@ def is_constant_invariant(self):
 
     def get_constant_output_channels(self, input_channels: AbstractSet[ChannelID]) -> AbstractSet[ChannelID]:
         return frozenset()
-
+    
+    @property
+    def contains_sweepval(self) -> bool:
```
shumpohl, line 69: Would it be a great churn to rename this to `contains_dynamic_value` and make it a method instead of a property?

shumpohl (changes requested):
Discuss replacement of sweepval with `dynamic_value`

qupulse/program/transformation.py
Nomos11: should be no problem.

what is the benefit of a method here over a property?

PR #921, opened 2025-07-30
other compatibility hacks remaining from experiment branch

Nomos11:


shumpohl:
Therefore, it is a proper hack

shumpohl:
Its less about nicer syntax and more about trying to keep the abstraction layer clean to make code easier to reason about.

PR #925, opened 2025-08-28
avoid string expr in to_next_multiple

Nomos11:
This is up to a factor 5-6 faster than before in the branch with min_quanta

### DIFF
--- a/qupulse/utils/__init__.py
+++ b/qupulse/utils/__init__.py
@@ -131,7 +131,7 @@ def forced_hash(obj) -> int:
 
 
 def to_next_multiple(sample_rate: ExpressionLike, quantum: int,
-                     min_quanta: Optional[int] = None) -> Callable[[ExpressionLike],ExpressionScalar]:
+                     min_quanta: Optional[int] = None) -> Callable[[ExpressionScalar], ExpressionLike]:
     """Construct a helper function to expand a duration to one corresponding to
     valid sample multiples according to the arguments given.
     Useful e.g. for PulseTemplate.pad_to's 'to_new_duration'-argument.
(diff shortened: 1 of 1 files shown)


qupulse/utils/__init__.py
```diff
@@ -151,6 +152,22 @@ def to_next_multiple(sample_rate: ExpressionLike, quantum: int,
...
-        #still return 0 if duration==0
-        return lambda duration: ExpressionScalar(f'{quantum}/({sample_rate})*Max({min_quanta},-(-({duration})*{sample_rate}//{quantum}))*Max(0, sign({duration}))')
-   
\ No newline at end of file
+        qI  = sp.Integer(quantum)
+        k   = qI / sample_rate  # factor to go from #quanta -> duration
+        mqI = sp.Integer(min_quanta)
+                
+        def _build_sym(d):
+            u  = d*sample_rate/qI                # "duration in quanta" (real)
```
shumpohl, line 160: Just use a descriptive variable name instead of a comment.

qupulse/utils/__init__.py
```diff
@@ -151,6 +152,22 @@ def to_next_multiple(sample_rate: ExpressionLike, quantum: int,
...
+            # max(mqI, ceil(u))  <=>  mqI if u <= mqI, else ceil(u)
+            # do not evaluate right now because parameters could still be variable,
+            # then it's just overhead.
+            return sp.Piecewise(
+                (0,  sp.Le(d, 0)),
+                (k*mqI, sp.Le(u, mqI)),
+                (k*ce, True)
+            , evaluate=False)
+        
+        return lambda duration: ExpressionScalar(_build_sym(duration))
```
shumpohl, line 173: Why not return _build_sym and put the ExpressionScalar in there?

Furthermore, for better performance it would make sense to construct as many expressions as possible at definition time?

qupulse/utils/__init__.py
```diff
@@ -151,6 +152,22 @@ def to_next_multiple(sample_rate: ExpressionLike, quantum: int,
...
+        mqI = sp.Integer(min_quanta)
+                
+        def _build_sym(d):
+            u  = d*sample_rate/qI                # "duration in quanta" (real)
+            ce = sp.ceiling(u)          # number of quanta after rounding up
+
+            # Enforce: 0 if d <= 0; else at least mqI quanta.
+            # max(mqI, ceil(u))  <=>  mqI if u <= mqI, else ceil(u)
+            # do not evaluate right now because parameters could still be variable,
+            # then it's just overhead.
```
shumpohl, line 166: But if the resulting expression ends up being used in multiple places it is a reasonable optimization. What did you profile?

shumpohl (changes requested):
I'll make my own version

qupulse/utils/__init__.py
shumpohl: But I think the duration of pt2 itself is used multiple times.

Nomos11: your branch is, for me:
construction:
```
%timeit pt.pad_to(pad_arg)
254 μs ± 1.76 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```
evaluation:
```
%timeit pt2.duration.evaluate_in_scope({'t_gate': 10.})
35.9 μs ± 174 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

Nomos11: a version mimicing the current one,
```
return (quantum / sample_rate) \
                    * sp.Max(min_quanta, sp.ceiling(d * sample_rate / quantum)) \
                    * sp.Max(0, sp.sign(d))

```
gives 900 us and 13.5 us, respectively.

Nomos11:
> At some point it probably makes more sense to change the expression backend

the next limiting factor would probably be `.subs` for pulses with heavy MappingPT usage. Sympy itself recognizes this is slow as an open issue https://github.com/sympy/sympy/issues/22240

shumpohl:
Ideally, we stir in a direction where the evaluations are not done at all, i.e. loop bodies are handled symbolically.

qupulse/utils/__init__.py
shumpohl: Looks good, or?

PR #933, opened 2026-01-26
Refactor context management for program builders

shumpohl:
This PR replaces `PulseTemplate._create_program(**a_lot_of__kwargs)` with `PulseTemplate._build_program(program_builder)`.

It moves all the information about the current scope, transformations, channel and measurement mappings into the program builder.

 - Add `ProgramBuilder.build_context: BuildContext`: A dataclass with mutable information like the current scope
 - Add `ProgramBuilder.build_settings: BuildSettings`: A dataclass with constant information like which templates to translate into a single waveform
 - Add `Program.get_defined_channels`.
 - Add `ProgramBuilder.override` to modify the current built settings/context.
 - Add `ProgramBuilder.with_mappings` for `MappingPT`
 - Add `ProgramBuilder.with_transformation` for arithmetic and channel modification PTs
 - Add `ProgramBuilder.with_metadata` for a semi standadized metadata passing. The only read metadata right now is the minimal sample rate.
 - Remove `ProgramBuilder.inner_scope` because it is no longer required.

To reduce code duplication the PR contains the `BaseProgramBuilder` class which contains some default implementations for context management which are shared between the loop and linspace builder.

TODO:

 - [x] fix other tests
 - [x] motivate split between BuildSettings and BuildContext or merge them
 - [x] Transform the PR text into a newspiece

### DIFF
--- a/qupulse/program/loop.py
+++ b/qupulse/program/loop.py
@@ -3,7 +3,8 @@
 # SPDX-License-Identifier: LGPL-3.0-or-later
 
 """Program builder implementation that creates the legacy :py:class:`.Loop`."""
-
+import contextlib
+import copy
 import reprlib
 import warnings
 from collections import defaultdict, OrderedDict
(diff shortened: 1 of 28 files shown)


shumpohl:
@Nomos11 triggered by properly implementing minimal sample rate

qupulse/program/loop.py
```diff
@@ -839,21 +842,21 @@ def time_reversed(self) -> ContextManager['LoopBuilder']:
             self._try_append(inner_program, None)
 
     @contextmanager
-    def with_sequence(self, measurements: Optional[Sequence[MeasurementWindow]] = None) -> ContextManager['ProgramBuilder']:
```
Nomos11, line 842: Was there a reason for removing the type hints?

shumpohl: They where wrong due to the decorator. The decorated function is a `Generator`. I will adjust them.

PR #936, opened 2026-01-28
Change to_single_waveform handling

### LINKED ISSUE #935: PulseTemplate identity based translation settings clash with "new" equality and hash implementation
Before 9edd1397165c7a15e38343e4a7ca4f4b01dcd1bf,  b41130e1aa243d5b1ebbbc48411c430b0df7617b and others hashing and equality of pulse templates was only based on obejct identity.

The `PulseTemplate.create_program` argument `to_single_waveform` was designed with that in mind to target specific pulse templates in a pulse hirachy. With the above commits this behavior changed and all equal pulse templates get the same treatment even if used in a different context.

The new `PulseTemplate.metadata` attribute allows setting translation settings per pulse template which is the replacement for the above usecase.

We should modify `to_single_waveform` to only accept pulse identifiers again to avoid surprising behaviour.

shumpohl:
Move to_single_waveform from PulseTemplate._build_program to ProgramBuilder.with_metadata and remove non-str member support.

Closes #935

### DIFF
--- a/changes.d/935.removal
+++ b/changes.d/935.removal
@@ -1,0 +1,2 @@
+The keyword argument `to_single_waveform` of `PulseTemplate.create_waveform` no longer accepts pulse template objects, only identifiers.
+Use `PulseTemplate.metadata.to_single_waveform` to enforce single waveform translation.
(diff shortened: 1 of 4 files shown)


shumpohl:
@Nomos11 to_stepping_repeat should land in the new `with_metadata` and follow the example implementation

